PR #4, opened 2019-09-11
Disable sched_setaffinity() under MacOS

aferrero2707:
I have added an ifdef to disable the code that calls sched_setaffinity() under MacOS, as this functionality is not present.


hanatos:
oh, just seeing this now! thanks, looking good. would you mind squashing the two commits so we wont have one in the history that doesn't compile?

hanatos:
..forget it. who needs clean git history at this stage.

PR #7, opened 2019-10-23
qvk: Add error checks to prevent segfaults

cryptomilk:


src/qvk/qvk.c
```diff
@@ -702,11 +702,27 @@ qvk_init()
   };
 
   /* create device and queue */
-  QVK(vkCreateDevice(qvk.physical_device, &dev_create_info, NULL, &qvk.device));
+  int res = vkCreateDevice(qvk.physical_device, &dev_create_info, NULL, &qvk.device);
```
hanatos: for this lengthy dance, there's the QVKR() macro

cryptomilk: From experience it is a bad idea to do a return in a macro. We had trouble in Samba with this and don't allow it anymore. But I can do that if you want.

cryptomilk: a) is the return value in the macro the correct one for all situation where the macro is called?
b) someone might not assume that it returns

hanatos: replace one line of code with 5 same lines of boilerplate code in many places? i don't think that helps legibility. isn't code duplication also one of the things that are written in books to not do? (if you're worried about this macro you'd love the doubly linked list..)

src/qvk/qvk.c
```diff
@@ -719,9 +719,9 @@ QVK_FEATURE_DO(inheritedQueries)
   /* create device and queue */
   QVK(vkCreateDevice(qvk.physical_device, &dev_create_info, NULL, &qvk.device));
 
-  vkGetDeviceQueue(qvk.device, qvk.queue_idx_graphics, 0, &qvk.queue_graphics);
-  vkGetDeviceQueue(qvk.device, qvk.queue_idx_compute,  0, &qvk.queue_compute);
-  vkGetDeviceQueue(qvk.device, qvk.queue_idx_transfer, 0, &qvk.queue_transfer);
+  QVK(vkGetDeviceQueue(qvk.device, qvk.queue_idx_graphics, 0, &qvk.queue_graphics));
```
hanatos: ..but this function does not return VkResult at all.

src/qvk/qvk.h
```diff
@@ -38,6 +38,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
     VkResult _res = __VA_ARGS__; \
```
hanatos: there is QVKR() for this purpose.

hanatos:
closing this, it doesn't seem to do anything at this point, let's wait for the underlying issue with AMD vulkan to progress. pushed a few early-out-return-error statements to qvk_init() in the meantime, in the hope to not completely destroy drivers when things go wrong.

PR #8, opened 2020-05-24
Zoom with mouse wheel scroll

phweyland:
Change in that PR: zoom on mouse wheel scroll

### DIFF
--- a/src/gui/main.c
+++ b/src/gui/main.c
@@ -84,6 +84,7 @@ char_callback(GLFWwindow* window, unsigned int c)
 static void
 scroll_callback(GLFWwindow *window, double xoff, double yoff)
 {
+  dt_view_mouse_scrolled(window, xoff, yoff);
   dt_gui_imgui_scrolled(window, xoff, yoff);
 }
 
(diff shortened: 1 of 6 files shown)


src/gui/main.c
```diff
@@ -84,7 +84,7 @@ char_callback(GLFWwindow* window, unsigned int c)
 static void
 scroll_callback(GLFWwindow *window, double xoff, double yoff)
 {
-  dt_gui_imgui_scrolled(window, xoff, yoff);
```
hanatos, line 87: is this not called any more now? i suppose this causes trouble for std imgui widgets that interact via mouse wheel? does for instance light table still scroll down with this line removed?

i suppose we should just call both. there is a dt_gui_imgui_want_mouse() function to wrap imgui's mouse request, but it seems to be always on..

hanatos:
nice! one question about the first bullet: did it not zoom centered around the mouse cursor before? you mean it centers the zoomed-in pixels under the mouse cursor instead of in the center of the screen?

src/gui/main.c
phweyland: > does for instance light table still scroll down with this line removed?

You are right. It doesn't work anymore.
Just added the line again.

phweyland:
> did it not zoom centered around the mouse cursor before?

Just checked. It did actually. My brain ...

src/gui/main.c
phweyland: > i suppose we should just call both. there is a dt_gui_imgui_want_mouse() function to wrap imgui's mouse request, but it seems to be always on..

I haven't tested that (as it was commented in your code).

PR #9, opened 2020-05-25
i-raw: Fix buffer size

cryptomilk:
```
pipe/modules/i-raw/main.cc:458:42: warning: ‘snprintf’ output may be truncated before the last format character [-Wformat-truncation=]
  458 |     snprintf(tmpfn, sizeof(tmpfn), "%s/%s", mod->graph->searchpath, fname);
      |                                          ^
pipe/modules/i-raw/main.cc:458:13: note: ‘snprintf’ output 2 or more bytes (assuming 513) into a destination of size 512
  458 |     snprintf(tmpfn, sizeof(tmpfn), "%s/%s", mod->graph->searchpath, fname);
      |     ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

mod->graph->searchpath is size 512, so the destination buffer needs to
be bigger.

### DIFF
--- a/src/pipe/modules/i-raw/main.cc
+++ b/src/pipe/modules/i-raw/main.cc
@@ -210,7 +210,7 @@ void modify_roi_out(
   // load image if not happened yet
   const char *fname = dt_module_param_string(mod, 0);
   const char *filename = fname;
-  char tmpfn[512];
+  char tmpfn[1024];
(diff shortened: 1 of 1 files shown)


hanatos:
correct, thanks for fixing.

PR #10, opened 2020-05-25
add histogram 'parade' ?

phweyland:
The changes in that PR were to make an exercise on connectors.
But it also makes a change on something I may have not understood: what does happen if there is more than 1024 points on the intensity in collect.comp ? It seems to me that rgb counts are mixed up. Would that be correct ?
I've also scaled a bit differently the output levels.
EDIT. Of course the UI input should be better with some check boxes (or combo box).

### DIFF
--- a/src/pipe/modules/hist/collect.comp
+++ b/src/pipe/modules/hist/collect.comp
@@ -16,9 +16,18 @@ layout( // input f16 buffer rgb
     set = 1, binding = 0
 ) uniform sampler2D img_in;
 
-layout( // output ui32 buffer rgb
+layout( // output ui32 buffer r
     set = 1, binding = 1, r32ui
-) uniform uimage2D img_out;
+) uniform uimage2D img_out_r;
+
+layout( // output ui32 buffer r
+   set = 1, binding = 2, r32ui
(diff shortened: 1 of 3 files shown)


hanatos:
yes, correct. this will overflow at some point. i mean.. as will the 32-bit per channel ints, albeit a whole lot later. i suppose splitting the buffers is the cleaner approach, maybe even with array connectors. but did you observe a difference between the approaches? did you run -d perf to see if it's more expensive? my guess is that more memory traffic might mean more cost.

as to ui: i suppose at some point we'd need a separate widget for the histogram (much like the thumbnail widget https://github.com/hanatos/vkdt/blob/master/src/gui/widget_thumbnail.hh ). for preferences like these we'll also likely need a config file in the future.

phweyland:
> did you run -d perf to see if it's more expensive?
A lot more expensive...
For hist collect, 
with your code: between 13 and 18 msec
with this code: between 54 and 65 msec.

hanatos:
okay thanks for checking.. not willing to pay so much for a histogram. closing this now, let's do ui improvements/parameters in a histogram widget separately.

PR #11, opened 2020-05-28
add boolean parameter

phweyland:
This PR may not be worth to be merged but this is also to get your advice / ideas about some related topics.

Boolean. I've used Checkbox(). Imgui offers also CheckboxFlags() but it is not straight forward to make correspond several ui booleans to a single module parameter (int).

Combo box. That's another one I miss. I think this should not be complicated to add it to your code. Obvious on params size, just to be able to read several values on params.ui side.
~~But these are strings. Is that ok for you to decode strings inside modules (.comp) ? Or do you have any other idea ?~~ EDIT combo() returns an int... Should be applicable also to checkboxflags() then.

I miss also two basic features on modules ui: reset and bypass. How do you see the second one ?

Having a draft of channel mixer to test this, it could be also useful to implement tabs definition in params.ui. Any thoughts ?

### DIFF
--- a/src/gui/render.cc
+++ b/src/gui/render.cc
@@ -452,7 +452,7 @@ void render_lighttable()
     ImGui::SetNextWindowPos (ImVec2(qvk.win_width - vkdt.state.panel_wd, 0),    ImGuiCond_Always);
     ImGui::SetNextWindowSize(ImVec2(vkdt.state.panel_wd, vkdt.state.panel_ht), ImGuiCond_Always);
     ImGui::Begin("panel-right", 0, window_flags);
-    
+
     float lineht = ImGui::GetTextLineHeight();
     float bwd = 0.5f;
     ImVec2 size(bwd*vkdt.state.panel_wd, 1.6*lineht);
@@ -760,7 +760,7 @@ inline void draw_widget(int modid, int parid)
       // TODO: distinguish by count:
       if(vkdt.graph_dev.module[modid].so->param[parid]->type == dt_token("float"))
       {
-        float *val = (float*)(vkdt.graph_dev.module[modid].param + 
+        float *val = (float*)(vkdt.graph_dev.module[modid].param +
           vkdt.graph_dev.module[modid].so->param[parid]->offset);
         char str[10] = {0};
         memcpy(str,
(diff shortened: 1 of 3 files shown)


phweyland:
Reviewing the code I've seen that it is useless to add a boolean parameter type, int is enough.
Added checkboxflags and combobox ui parts.
Both need a list of entries in params.ui
In params.ui we have these entries (sort of):
```
preslum:checkbox
grey:checkbox
flags:checkbxf:preslum1:grey1:red1:green1:blue1
channel:combobox:red:green:blue:grey
```

while we have in params:
```
preslum:int:1:0
grey:int:1:0
flags:int:1:0
channel:int:1:0
```

Is that coherent with your plans ? 
Do we need to consider these list items as tokens too ? (for this we miss  a way to get them without knowing their number)
Do have already ideas to translate the tokens into human language ?

PS: Please do not consider too much my channel mixer code. I know it is not correct... :(

hanatos:
many questions! let me try to answer a few:

yes, i agree an int is enough (all c knows, too) for boolean types, as well as combobox entries. this means we'll have to be consistent with regard to the meaning/ordering of combo box entries in the future. module versioning is another thing that is in the concept of the pipe but not implemented..

by ui reset you mean the whole module? i was meaning to implement a slider widget that accepts double clicks in the future.

i haven't thought about bypass (temporarily disable) modules yet. probably i would just remove it from the graph (easy, there's a button already) and then put it back (slightly harder, the module will need to remember where it's been the last time around, potentially not by moduleid but by name/instance such that it can be put back even if the rest of the graph changed). it will not be simple to create a generic "bypass" option for all modules, since not all of them have a single input and a single output. say a module creates a highpass and lowpass from the input, what would be the bypass in this case. i'm currently checking for inputs and outputs named "input" and "output" to offer some ui shortcut for frequent cases. this would be useful in this case, too.

your proposal for params and params.ui look good. the list items don't need to be tokens, that's just for stuff that potentially needs to be sorted and compared very quickly, as well as potentially be put into a binary structure for fast parsing.

i also haven't thought about tabs in the ui. the params.ui file format is very rudimentary anyways. we could probably append a tag like "tab0" or "tab1". this would then be different in params.ui (the module) and darkroom.ui (the collection of favourite gui elements in the first tab).

translation: probably intltool style translation can be built into this using a list of N_() and then only translating things on demand when passing it to the imgui widgets. this is slightly dangerous as some imgui event handlers make use of the widget name if not explicitly passing an ID (internally, not my code).

hanatos:
closing this now in an effort to try and tidy up the pull requests while there are so few that i still have a chance to do this.

re: booleans and checkboxes: i like the strategy we had in the early darktable days: don't use bool and checkbox. use ints (c style) and use combo boxes (supported in vkdt now), because these allow us to put more speaking text to each option. this can sometimes help a lot when you aren't sure whether the checkbox indicates "on" or "off" in some colour scheme, and then whatever it means to switch, say "mode on" or "mode off". one such example might be the current colour module, which has a "parametric" and "data driven" mode, which might otherwise be a boolean, too.

PR #12, opened 2020-06-02
vkdt specific slider widget

phweyland:
A try to make mouse wheel work on sliders. Could work the same way for combo boxes. 

Some doubts.
The easy one. I've added some data to vkdt (dt_gui_t) but I think this is not the best place. You have probably a better idea.

The strange ones. For the last 3 or 4 days I've had different and moving seg faults, apparently not related to the changes I had made. One of the most frequent was in line 335 in thumbnails.c. Finally I've got some peace, with 3 changes:
- make sure that vulkansdk is up and running for validation layers
- set line 448 in qvk.c: `float queue_priorities[10] = {1.0f};` as I got a kronos validation error related to invalid priority (on an index <> 0 if I remember well).
- set line 22 in gui/main.c as `dt_gui_t vkdt = {0};`, still is that PR.

Very often, deleting all .o files before building made it work. but not always.

As usual any suggestion is welcome.

### DIFF
--- a/src/gui/render.cc
+++ b/src/gui/render.cc
@@ -835,8 +835,8 @@ inline void draw_widget(int modid, int parid)
         float oldval = *val;
         char str[10] = {0};
         memcpy(str, &param->name, 8);
-        if(ImGui::dt_SliderFloat(str, val,
-              param->widget.min, param->widget.max, "%2.5f"))
+        if(ImGui::DtSliderFloat(str, val,
+              param->widget.min, param->widget.max, param->val[0], "%2.5f"))
         {
           dt_graph_run_t flags = s_graph_run_none;
           if(vkdt.graph_dev.module[modid].so->check_params)
@@ -852,8 +852,8 @@ inline void draw_widget(int modid, int parid)
         int32_t oldval = *val;
         char str[10] = {0};
         memcpy(str, &param->name, 8);
-        if(ImGui::dt_SliderInt(str, val,
-              param->widget.min, param->widget.max, "%d"))
+        if(ImGui::DtSliderInt(str, val,
+              param->widget.min, param->widget.max, param->vali[0], "%d"))
         {
           dt_graph_run_t flags = s_graph_run_none;
           if(vkdt.graph_dev.module[modid].so->check_params)
--- a/src/gui/render.cc
+++ b/src/gui/render.cc
@@ -9,6 +9,7 @@ extern "C" {
 extern int g_busy;  // when does gui go idle. this is terrible, should put it in vkdt.gui_busy properly.
 }
 #include "gui/widget_thumbnail.hh"
+#include "gui/widget_slider.hh"
 #include "imgui.h"
 #include "imgui_impl_vulkan.h"
 #include "imgui_impl_glfw.h"
@@ -834,7 +835,7 @@ inline void draw_widget(int modid, int parid)
         float oldval = *val;
         char str[10] = {0};
         memcpy(str, &param->name, 8);
-        if(ImGui::SliderFloat(str, val,
+        if(ImGui::dt_SliderFloat(str, val,
               param->widget.min, param->widget.max, "%2.5f"))
         {
           dt_graph_run_t flags = s_graph_run_none;
@@ -851,7 +852,7 @@ inline void draw_widget(int modid, int parid)
         int32_t oldval = *val;
         char str[10] = {0};
         memcpy(str, &param->name, 8);
-        if(ImGui::SliderInt(str, val,
+        if(ImGui::dt_SliderInt(str, val,
               param->widget.min, param->widget.max, "%d"))
         {
           dt_graph_run_t flags = s_graph_run_none;
@@ -1264,7 +1265,7 @@ void render_darkroom()
         case dt_token("crop"):
         {
           float v[8] = {
-            vkdt.wstate.state[0], vkdt.wstate.state[2], vkdt.wstate.state[1], vkdt.wstate.state[2], 
+            vkdt.wstate.state[0], vkdt.wstate.state[2], vkdt.wstate.state[1], vkdt.wstate.state[2],
             vkdt.wstate.state[1], vkdt.wstate.state[3], vkdt.wstate.state[0], vkdt.wstate.state[3]
           };
           float p[8];
@@ -1277,7 +1278,7 @@ void render_darkroom()
         case dt_token("pick"):
         {
           float v[8] = {
-            vkdt.wstate.state[0], vkdt.wstate.state[2], vkdt.wstate.state[1], vkdt.wstate.state[2], 
+            vkdt.wstate.state[0], vkdt.wstate.state[2], vkdt.wstate.state[1], vkdt.wstate.state[2],
             vkdt.wstate.state[1], vkdt.wstate.state[3], vkdt.wstate.state[0], vkdt.wstate.state[3]
           };
           float p[8];
--- a/src/gui/widget_slider.hh
+++ b/src/gui/widget_slider.hh
@@ -1,0 +1,299 @@
+#pragma once
+#include "imgui.h"
+
+namespace ImGui {
+
+  // Those MIN/MAX values are not define because we need to point to them
+  static const ImS32          IM_S32_MIN = INT_MIN;    // (-2147483647 - 1), (0x80000000);
+  static const ImS32          IM_S32_MAX = INT_MAX;    // (2147483647), (0x7FFFFFFF)
+
+  template<typename TYPE, typename FLOATTYPE>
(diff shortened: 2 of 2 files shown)


phweyland:
It would be more elegant to have this feature native in imgui.
https://github.com/ocornut/imgui/pull/3281

hanatos:
interesting, i don't get the validation layer message nor the segfaults. will try to reproduce and keep an eye open. possible that some struct size is inlined in an old .o file and causes trouble. good catch with the queue_priorities anyways, this is clearly too short for three queues.. also initing the global struct to zero cannot hurt, let me find a place other than the main gui function (so the cli would not run into the same thing later).

hanatos:
re: mouse wheel in the widget: i think we may need our own widget anyways (double click to reset, maybe a detail/zoom mode as dt has it). it's easy enough to do with the drawing primitives imgui provides.

phweyland:
> re: mouse wheel in the widget: i think we may need our own widget anyways (double click to reset, maybe a detail/zoom mode as dt has it). it's easy enough to do with the drawing primitives imgui provides.

Ok, I'll try to make something like in https://github.com/hanatos/vkdt/blob/master/src/gui/widget_thumbnail.hh as you suggested.

phweyland:
Sorry if my question is a bit stupid ... I have not found the way to override the definition of ImGui::SliderBehaviorT(). So I don't see any other way than rewrite SliderFloat() from scratch into vkdt. Is that your view too ? If not, any tip about the best way to follow ?

hanatos:
let's see if i understand correctly. you would rewrite another function,
right? like DtSliderFloat() or whatever and then use/replicate as much as
you can from imgui's slider implementation? i don't think we're using
SliderFloat in many places, all of the ui is pretty much automatically
generated from config files. should be very easy to drop in our own call. i
would certainly leave the original imgui functions as they are.

phweyland:
Not sure to have understood perfectly your suggestion... Here is the try.
I've copied the imgui pieces of code for imgui slider, removed the power part and added the mouse wheel control (+ Ctrl & Shift dt modifier).
Remains probably some cleaning to do. Do we need to keep the keyboard navigation (seems to be deactivated currently) ?
If you find that way convenient I'll try to add the slider reset.

hanatos:
sorry for the late reply, got distracted and then i lost track because i was confused about the PR name :)

yes, i think we should do it exactly like this.

minor remark:
as much as i hate camel case.. we should probably do it consistently and then call it DtSliderFloat or dt_slider_float. the c++/imgui side of code is already cameled a lot, so i wouldn't mind going DtSliderFloat here.

this is currently only replicating the exact imgui behaviour, right? i didn't try to build/run this yet.

phweyland:
> this is currently only replicating the exact imgui behaviour, right?

Imgui behaviour plus scroll wheel.

Is missing the reset function. Here is the question where to pick up the default values ? Image cfg is not the right source. default-darkroom.cfg may not have all the values. A specific default.cfg may be necessary. What do you think ?



> DtSliderFloat or dt_slider_float

I'm fine with both.
I'll rebase and resolve conflicts.

phweyland:
> I'll rebase and resolve conflicts.

new version of vulkan...  VK_API_VERSION_1_2
On debian buster I've only the version 1.1.  I've made a tried of installing libvulkan 1.2 but there are unmet dependencies... risky ?
 

> Is missing the reset function
Resolving the rebase conflicts I've seen you have made something on this.

hanatos:
great, thanks for rebasing. re: defaults: the modules store default values
that come out of the 'params' files. it's part of the param struct in
module->so. this is the global struct, the class type of a module if you
will.

https://github.com/hanatos/vkdt/blob/master/src/pipe/global.h#L66

https://github.com/hanatos/vkdt/blob/master/src/pipe/params.h#L18

for instance this is how defaults are inited when adding a new module to
the graph:

https://github.com/hanatos/vkdt/blob/master/src/pipe/module.c#L47

(you'd only be interested in the reading part)

hope that helps.

hanatos:
oh, right. if you only have older vk, you can roll back
a1a226aeb3b25066d71721ede884c44dfbfa8bd4 and i suppose probably also
137a237d2db78daea0bc8e53e0440baec3aa7696 because it requires a glsl
compiler with vk 1.2 support. i guess i was a bit too happy to find it in
my debian sources on that day..

phweyland:
> [a1a226a](https://github.com/hanatos/vkdt/commit/a1a226aeb3b25066d71721ede884c44dfbfa8bd4) and i suppose probably also [137a237](https://github.com/hanatos/vkdt/commit/137a237d2db78daea0bc8e53e0440baec3aa7696) because it requires a glsl

Great! I can build again now.

phweyland:
A bit of gymnastic but ... voilà. 
Sometimes the double click is followed by a click ... Do you see an alternative to double-click ?

hanatos:
hi, just a quick heads up that i'm trying to test this :)

i'm getting this:

```
clang gui/gui.o gui/render.o gui/darkroom.o gui/main.o gui/view.o ../ext/imgui/imgui.o ../ext/imgui/imgui_draw.o ../ext/imgui/imgui_widgets.o ../ext/imgui/examples/imgui_impl_vulkan.o ../ext/imgui/examples/imgui_impl_glfw.o qvk/qvk.o qvk/qvk_util.o core/log.o core/threads.o pipe/alloc.o pipe/connector.o pipe/global.o pipe/graph.o pipe/graph-io.o pipe/graph-export.o pipe/module.o db/db.o db/thumbnails.o -o vkdt \
     -lvulkan -ldl -lglfw -lm -lstdc++ -lfreetype -ldl -pthread  
/usr/bin/ld: gui/render.o: in function `bool ImGui::SliderBehaviorT2<int, int, float>(ImRect const&, unsigned int, int, int*, int, int, int, char const*, ImRect*)':
/home/jo/vc/vkdt/src/./gui/widget_slider.hh:169: undefined reference to `int ImGui::RoundScalarWithFormatT<int, int>(char const*, int, int)'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
as a linker error when building. i think maybe to avoid duplicating *all* this code we should be using the IMGUI_INCLUDE_IMGUI_USER_INL and IMGUI_INCLUDE_IMGUI_USER_H features to inject our widgets into the same compilation unit and namespace as imgui, so we can reuse some of the private features. also maybe it's possible to drag out the

```
if (g.HoveredId == id && g.IO.MouseDoubleClicked[0])
```

test *outside* the SliderBehaviour function, so we can just call into the regular code in case it's not a double click? this might avoid the need to call so many internal functions (and hence the IMGUI_USER defines above) in the first place?

phweyland:
> also maybe it's possible to drag out the
> 
> ```
> if (g.HoveredId == id && g.IO.MouseDoubleClicked[0])
> ```

I've made a try on this. But before SliderFloat(), the widget is not drawn and the IsItemHovered() is not related to the right widget. After SliderFloat() the widget is already drawn with the new value.

> using the IMGUI_INCLUDE_IMGUI_USER_INL and IMGUI_INCLUDE_IMGUI_USER_H features

If I understand correctly it is a way to put some user code inside imgui package (in the imgui folders). Is that correct ? Is it what you suggest ?

I think I could simplify significantly the code and avoid the conflicting dependencies.

hanatos:
re: IMGUI_USER stuff. i just pushed my half assed attempt to the
imgui-widgets branch. this is only using the USER_H feature, not the
USER_INL. but maybe you'll see the idea? the USER_INL macro will include an
.inl file similar to the header for the other macro. we could have a
imgui_user.inl or what it's called including the inlined .cc files with the
meat for the widgets.

note that this requires only to define the macros during compilation (see
flat.mk), and to put our files in the search path. i.e. the imgui
repository is unchanged.

the version i pushed unfortunately has the same compilation error as your
branch had on my system, so i don't know if it would be functional
otherwise.

phweyland:
I've built your branch imgui-widgets, but there is conflict here:
`g++ -Wall -pipe -I. -D_GNU_SOURCE -Wall -pipe -O3 -march=native -g  -DIMGUI_INCLUDE_IMGUI_USER_H -Igui/widgets/ -I../ext/imgui -I../ext/imgui/examples/ -c ../ext/imgui/imgui.cpp -o ../ext/imgui/imgui.o`

-DIMGUI_INCLUDE_IMGUI_USER_H shouldn't be used to compile imgui.cpp. 
Removing it let me build vkdt.
As I'm unknowledgeable about make ... I let you fix it.

hanatos:
i think this is now outdated/conflicted enough to close it.

in the mean time i implemented double-click-resets for most sliders, so immediate changes are less pressing workflow wise. i'm leaving a more comprehensive gui/workflow overhaul for a later release.

PR #13, opened 2020-09-02
reformatting

aurelienpierre:
This is only my IDE being pedantic with auto-formatting. The style is supposed to be compliant with darktable, not sure which one you want to use here. I'm just dropping the format changes in an early PR before starting to edit these files to avoid massively noisy commits.

### DIFF
--- a/src/gui/gui.c
+++ b/src/gui/gui.c
@@ -15,38 +15,38 @@
 static void
 style_to_state()
 {
-  const float pwd = vkdt.style.panel_width_frac * (16.0/9.0) * qvk.win_height;
-  vkdt.state = (dt_gui_state_t) {
-    .look_at_x = FLT_MAX,
-    .look_at_y = FLT_MAX,
-    .scale = -1.0f,
-    .center_x = vkdt.style.border_frac * qvk.win_width,
-    .center_y = vkdt.style.border_frac * qvk.win_width,
-    .panel_wd = pwd,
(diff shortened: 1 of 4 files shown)


src/gui/gui.c
```diff
@@ -15,140 +15,139 @@
...
+        .flags = VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT,
     };
-    QVKR(vkCreateCommandPool(qvk.device, &cmd_pool_create_info, NULL, vkdt.command_pool+i));
+    QVKR(vkCreateCommandPool(qvk.device, &cmd_pool_create_info, NULL, vkdt.command_pool + i));
 
     VkCommandBufferAllocateInfo cmd_buf_alloc_info = {
-      .sType              = VK_STRUCTURE_TYPE_COMMAND_BUFFER_ALLOCATE_INFO,
-      .commandPool        = vkdt.command_pool[i],
-      .level              = VK_COMMAND_BUFFER_LEVEL_PRIMARY,
-      .commandBufferCount = 1,
```
hanatos, line 112: really? this is nonsense. it will just misalign the '='

i don't know why IDEs like to burn cpu cycles on this.

aurelienpierre:
Mind you, I had to install an extension to get alignment on "=". Otherwise I forced it to LLVM flavor of clang-format.

hanatos:
sorry will not re-indent because a random microsoft tool says so. can't it just leave the code the dev doesn't touch as it was?

PR #37, opened 2022-06-09
load_directory(tags): doesn't show deleted images

### LINKED ISSUE #36: vkdt hangs on trying to open a non existing (deleted) image from tag collection
to reproduce:
- tag some images (with or without *cfg) as "x"
- delete one of these images
- go to "x" collection.
- try to open the deleted image

In darkroom_enter(), changing `char graph_cfg[PATH_MAX+100];` to  `char graph_cfg[PATH_MAX+100] = {0};` avoids hanging but the issue may be deeper.
`dt_db_load_directory()` should check if the image exists or/and when deleting the image, the corresponding symlinks should be removed.

Related somehow, deleting an image in tag collection doesn't delete the image (only the symlink).

EDIT: That's a reason to be able to get the tags per image...

phweyland:
attempt to fix #36

workaround waiting for tags folder proper clean up.

### DIFF
--- a/src/db/db.c
+++ b/src/db/db.c
@@ -230,7 +230,23 @@ void dt_db_load_directory(
           continue;
         }
       }
-      else ep_len -= 4; // remove '.cfg' suffix
+      else
+      {
+        ep_len -= 4; // remove '.cfg' suffix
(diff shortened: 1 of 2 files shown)


hanatos:
i think #36 is fixed, thus closing this pr now.

PR #38, opened 2022-06-20
tags: add image's tag list

phweyland:
Started to add the image's list of tags.
The attach part is more or less done.
The detach part is started but... 
First stringpool has no free feature. Attaching detaching a tag too many times would fulfill the list. Maybe not that important as it's reset at any change of collection.
Then, currently there is no collection flag to say if it's folder or tag collection. Searching where this should be done. Knowing that would also allow to display "detach image" instead of "delete image" accordingly.
Am I on a valid path ?

### DIFF
--- a/src/db/db.c
+++ b/src/db/db.c
@@ -35,6 +35,7 @@ dt_db_cleanup(dt_db_t *db)
   // do not write if opened single image:
   if(db->image_cnt > 1) dt_db_write(db, dbname, 0);
   dt_stringpool_cleanup(&db->sp_filename);
+  dt_stringpool_cleanup(&db->sp_tags);
   free(db->collection);
   free(db->selection);
   free(db->image);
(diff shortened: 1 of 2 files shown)


phweyland:
> Am I on a valid path ?

I'm afraid not. Detaching (deleting) an image symlink, the corresponding real image's folder collection is not in memory. So there is no direct way to update the tags attached to an image. 
It would be necessary to get the real symlink path, extract the dirname, and update the vkdt.db directly...
Any other suggestion ?

hanatos:
all that sounds valid to me. the way to attach a tag looks straight forward and i think i would have done it in a very similar way.

detaching a tag: yes, from image to tag symlink means compute the hash on the filename, the other way it's realpath(). deleting the symlink would now have to be done in lockstep with deleting the entry in the vkdt.db file. that does not sound efficient.

an alternative is: detach: just delete the link and accept inconsistent vkdt.db files in the source directories.

when opening a directory: perform a consistency pass/collect all tags by searching for the hashes in the tags subdirectories. also sounds inefficient.

i suppose this consistency pass would be needed at some point anyways because storing the same data in two places will necessarily break (if for nothing else then while debugging or after crashes).

maybe there's a more efficient way of deleting the tag from vkdt.db or a similar file? i don't think i have good answers here.

hanatos:
closing this as stale now..

PR #44, opened 2022-07-05
vkdt-cli: add i-format option

phweyland:
to let export non raw files through vkdt-cli

### DIFF
--- a/src/cli/main.c
+++ b/src/cli/main.c
@@ -32,6 +32,8 @@ int main(int argc, char *argv[])
   {
     if(!strcmp(argv[i], "-g") && i < argc-1)
       param.p_cfgfile = argv[++i];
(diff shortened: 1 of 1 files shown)


hanatos:
wait what are you trying to do there? and why are you not passing a cfg file to the cli? this looks wrong to me but also i don't understand what you're trying to do here.

phweyland:
I just want to create thumbnails with vkdt-cli. That works implicitly for raw (in graph_export I-raw is set by default) but jpg are ignored. Then vkdt-cli needs a way to know that the image is a jpg.

phweyland:
> i'm guessing that you call `vkdt-cli -g something.jpg.cfg` where the cfg
does not in fact exist

Exactly, I've missed to mention this.

> could you try whether that works 

Currently a bit far of the computer, I'll try this asap. 
Thanks!

PR #83, opened 2023-04-13
initial library, libinstall

hanatos:


Makefile
```diff
@@ -18,6 +18,8 @@ all: ext src bin
 prefix?=/usr
 DESTDIR?=
 VKDTDIR?=$(DESTDIR)$(prefix)/lib/vkdt
+VKDTLIBDIR=/usr/local/lib/vkdt
```
hanatos, line 21: probably better to stick to $(DESTDIR)$(prefix) so packagers can drop their temp dirs and put `/usr/local/` in $(prefix) if they want.

Makefile
```diff
@@ -34,6 +36,28 @@ install: all
 	ln -rsf ${VKDTDIR}/vkdt $(DESTDIR)$(prefix)/bin/vkdt
 	ln -rsf ${VKDTDIR}/vkdt-cli $(DESTDIR)$(prefix)/bin/vkdt-cli
 
+libinstall: ext lib bin
+	mkdir -p $(VKDTLIBDIR)
```
hanatos, line 40: this one is not necessary because `mkdir -p` will make the parent too

Makefile
```diff
@@ -34,6 +36,28 @@ install: all
...
+	cp -rfL bin/data ${VKDTLIBDIR}
+	rsync -avP --include='**/params' --include='**/connectors' --include='**/*.ui' --include='**/ptooltips' --include='**/ctooltips' --include='**/readme.md' --include='**.spv' --include='**.so' --include '*/' --exclude='**' bin/modules/ ${VKDTLIBDIR}/modules/
+	cp -rfL bin/libvkdt.so ${VKDTLIBDIR} 
+	cp -rfL bin/libvkdt.h $(VKDTINCDIR)
+	cp -rfL src/qvk/*.h $(VKDTINCDIR)/qvk
+	cp -rfL src/pipe/*.h $(VKDTINCDIR)/pipe
+	cp -rfL src/pipe/modules/*.h $(VKDTINCDIR)/pipe/modules
+	cp -rfL src/gui/*.h $(VKDTINCDIR)/gui
+	cp -rfL src/core/*.h $(VKDTINCDIR)/core
+	echo "$(VKDTLIBDIR)" > /etc/ld.so.conf.d/libvkdt.conf
```
hanatos, line 58: oh! root make rules! i'd probably either not do this at all (leave it to the user to either install to where ldd can find it or take care of this in a distro-specific way) or maybe for convenience provide this as a separate rule.

if people build packages, they would set DESTDIR to something say /tmp/package-version/ and do like a test install to grab all the installed files from later on, to put in the package. if the install make target also messes with the global system configuration this becomes messy.

Makefile
```diff
@@ -85,6 +114,12 @@ distclean:
 	rm -rf bin/modules
 	rm -rf src/macadam.lut
 
+libclean:
+	rm -rf $(VKDTLIBDIR)
+	rm -rf $(VKDTINCDIR)
+	rm -r /etc/ld.so.conf.d/libvkdt.conf
```
hanatos, line 120: see above.

src/core/fs.h
```diff
@@ -84,11 +90,23 @@ fs_basedir(
     size_t maxlen)  // allocation size
 {
 #ifdef __linux__
+  #ifdef LIBVKDT
```
hanatos, line 93: i think i'd never indent preprocessor macros, but no strong opinion.

src/core/fs.h
```diff
@@ -84,11 +90,23 @@ fs_basedir(
...
+  void *handle;
+  char mod[PATH_MAX];
+  handle = dlopen("libvkdt.so", RTLD_LAZY);
```
hanatos, line 96: are you leaking this? how does dlopen work? is it reference counted so that you can safely call dlclose and the main dynamic linkage will not disappear?

src/core/fs.h
```diff
@@ -84,11 +90,23 @@ fs_basedir(
...
+  if (!handle) {
+    fprintf(stderr, "%s\n", dlerror());
+	exit(1);
```
hanatos, line 99: ouch! i mean if we can't find the library then we're in deeper trouble than this :) but `exit()` is totally skipping any cleanup that may come in the main thread. i'm a bit afraid of what happens to gpus if you lock them up, and probably annoying if you're not writing config files or some such.

i'd just let the code fall through to whatever comes next and try to find a basedir the other way.

src/core/fs.h
```diff
@@ -84,11 +90,23 @@ fs_basedir(
...
+  }
+  dlinfo(handle, RTLD_DI_ORIGIN, &mod);
+  snprintf(basedir, maxlen, "%s/", mod);
```
hanatos, line 102: i think the `/` is double. it'll be used as `basedir + '/' + somesuch` later on.

src/lib/libvkdt.h
```diff
@@ -0,0 +1,23 @@
+#ifndef LIBVKDT_H
```
hanatos, line 1: the rest of the code base uses `#pragma once` instead.

src/lib/libvkdt.h
```diff
@@ -0,0 +1,23 @@
...
+#include "pipe/modules/api.h"
+#include "core/log.h"
+#include "core/version.h"
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
+
```
hanatos, line 23: the whole file is just a convenience thing? i suppose the cplusplus nonsense is useful to avoid a lot of typing, and also there may be a minimal necessary set of headers, so i guess this is fine.

Makefile
butcherg: Agree, but /usr/local contains lib, include, etc., not usr.

Makefile
butcherg: Agree.

Makefile
butcherg: I'm learning this as I go, but it looks to me ldconfig has to be involved in the dlopen/dlinfo dance below, and this is how it's set up.

Makefile
butcherg: +1

src/core/fs.h
butcherg: Yeah, in a hurry to post before bedtime, forgot to include a dlclose() call.

src/core/fs.h
butcherg: Indeed.  Need to think through other ways, and/or return a result code from fs_basedir().

src/core/fs.h
butcherg: The path returned by dlinfo() was being trounced by the call to fs_dirname() at the end of fs_basedir().  Small edit window on my tablet, scratched my head for a couple of hours trying to figure out why the path was losing its last member.  This was just a way to keep fs_dirname() in the same place...

src/lib/libvkdt.h
butcherg: Agree.

Makefile
hanatos: DESTDIR is only to create packages i think. so to install to /usr/local, you'd set `prefix=/usr/local` and leave DESTDIR empty.

Makefile
hanatos: well yes for the system to pick it up automatically you need to rerun ldconfig. that's done after installing packages, and people who install libraries to their system usually remember doing it i would assume. also the `/etc/ld.so.conf.d/` way of doing things may or may not be a universal thing on all distros. works on debian (tm).

i'm more worried about the separation of concerns here. again, maybe you just want to try an install to a non-zero DESTDIR to collect all files into a tarball. you don't want that install rule to modify your host system (especially not with a library path pointing to /tmp/something...).

src/core/fs.h
hanatos: aaah i see. because you're not straight away returning it if successful. okay then i suppose both works: '/' or return right after.

src/lib/libvkdt.h
hanatos: nah for a convenience thing leave the cplusplus guards in. as to api.. this probably needs careful consideration and probably internal restructuring too. some of the graph modification stuff is explicitly exposed to modules in `src/pipe/modules/api.h`. 

theu `ui_callback` is just a special case for a ui element that consists only of a button and triggers some code in the module. conceptually it's no more generic than a slider attached to a parameter. these ui hooks are listed here:
https://github.com/hanatos/vkdt/blob/master/src/pipe/global.c#L191

and yes, graph management is probably about to become more complex with the high level collapsible blocks with multiple inputs/outputs..

hanatos:
about `-fPIE` and `-fPIC`. `vkdt` uses the former, for a library build you'd need the latter. executables are built with `-pie`, so at the price some small overhead and at the cost of [allowing symbol interposition](https://www.drdobbs.com/building-library-interposers-for-fun-and/184404926), you'll probably want to replace

```
EXE_CFLAGS=-fPIE
```
by
```
EXE_CFLAGS?=-fPIC
```
where the question mark means to set it only in case a user didn't provide an override up front (say from `config.mk`).

butcherg:
I just pushed a new commit to the branch, tested installation/clean to /usr/local, and compilation/link/run of libtest.  User has to run sudo ldconfig after the install.

Thing is, as-is the libinstall target has to put data/ and modules/ in /usr/local/lib so the dlinfo() can provide a directory that resolves them.  Maybe some search strategy that allows a subdirectory in fs_basedir(), e.g.  first look in dlinfo_dir, then dlinfo_dir/vkdt?

hanatos:
merged!

and adjusted a few things:
- unified the install rules in the makefile a bit, reusing stuff
- the basdir is now something/lib/vkdt if the lib is in something/lib, which is the same as if the install includes binaries in something/lib/vkdt/vkdt-cli
- basedir is probed, it uses binary/exe and tests for `darkroom.ui`. if it can't find it, it reverts to searching `libvkdt.so`
- renamed `libvkdt.h` to `vkdt.h`, using lib in a header file name seemed duplicate.

i suppose up next: write test code using the library and put it into a github action to make sure this code path does not rot over time.

PR #92, opened 2023-10-19
Add support for full-range yuv video

paolodepetrillo:
The h264 .mov files from my Olympus E-M1 II, and probably some of other DSLR / mirrorless, use the full scale (0-255) yuv encoding. This patch rescales the values to the 16-235/240 range expected by the rgb conversion matrix.

A short sample .mov that I used for testing is attached:

https://github.com/hanatos/vkdt/assets/3238173/eeb2e434-7221-4384-be2d-8631844e63ed

### DIFF
--- a/src/pipe/modules/i-vid/conv.comp
+++ b/src/pipe/modules/i-vid/conv.comp
@@ -5,10 +5,11 @@
 layout(local_size_x = DT_LOCAL_SIZE_X, local_size_y = DT_LOCAL_SIZE_Y, local_size_z = 1) in;
 layout(std140, set = 0, binding = 1) uniform params_t
 {
-  int color; // bt.709, bt.2020
-  int trc;   // linear, bt.709, smpte2084, HLG
-  int bits;  // 8, 10, 12, 16
-  int chr;   // 420, 422, 444, we don't care here
+  int color;    // bt.709, bt.2020
+  int trc;      // linear, bt.709, smpte2084, HLG
+  int bits;     // 8, 10, 12, 16
+  int chr;      // 420, 422, 444, we don't care here
+  int colrange; // 16-235/240, 0-255
 } params;
 
 layout(set = 1, binding = 0) uniform sampler2D img_in[];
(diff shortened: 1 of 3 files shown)


hanatos:
heya, thanks for the pull request!

looking good, i think in the long run we'll not come around this extra parameter for range. in theory it would be possible to abuse the bits == 16 case i suppose, but i like your approach better.

maybe can you add the same case to the bt.2020 code path? i.e. if full range don't do the rescaling dance around here: https://github.com/hanatos/vkdt/blob/master/src/pipe/modules/i-vid/conv.comp#L36

also if you add a line of description to the readme https://github.com/hanatos/vkdt/blob/master/src/pipe/modules/i-vid/readme.md#parameters it will be parsed by the build system and appear as tooltip text in the ui.

hanatos:
merged manually with the changes described above.

paolodepetrillo:
I had been working on this yesterday, I think the conversion I had used was slightly incorrect (some off by one errors), and was also trying to simplify it by using the same de-quantization code for color==0 or color==1. I have not fully tested it for all cases yet but just pushed the latest code to my branch.

Also, for trc==1 should the bt709 TRC be used instead of the sRGB TRC?
`rgb = mix(pow((rgb+0.099)/(1.0+0.099), vec3(1.0/0.45)), rgb/4.5, lessThan(rgb, vec3(0.081)));`
instead of:
`rgb = mix(pow((rgb+0.055)/(1.0+0.055), vec3(2.4)), rgb/12.92, lessThanEqual(rgb, vec3(0.04045)));`

hanatos:
ah sorry then i pushed too early :)

maybe it would be a good idea to remove this composite matrix i found on
the internet and do it properly step by step? i.e. the full range code path
does nothing and the mpeg code path stretches out as needed after applying
the same/regular matrix?

oh, you're probably right about the trc, need to double check. it does
looks slightly off i thought.

paolodepetrillo:
I did remove that matrix from silicondust.com - now bt709 and bt2020 work exactly the same and the only difference is the coefficients of the yuv to rgb matrix. For the full and limited range paths I tried to exactly follow the formulas in ITU-T H.264 section E.2.1 to undo the quantization (resulting in Y from 0-1 and Cb/Cr from -0.5 to 0.5 in all cases), and then the yuv to rgb matrix is applied to that.

hanatos:
sounds great! can you do a new pull request?

PR #93, opened 2023-10-27
Fixes for i-vid inputs

paolodepetrillo:
Based on #92 this adds support for either full-range or limited-range video for any bit depth. It also changes the TRC from sRGB to bt709 when the file metadata indicates bt709. It has not been tested extensively for every possible combination of inputs but seems to work.

This was a useful reference, copies a lot of the formulas from the various ITU docs all in one place: https://registry.khronos.org/DataFormat/specs/1.3/dataformat.1.3.html#_introduction_to_color_conversions

This also uses a ui16 format input to the shader in all cases, even when it shouldn't be necessary because the video input is 8-bit. A version without that change is [here](https://github.com/paolodepetrillo/vkdt/tree/fix-video-color-range). This fixed a weird issue that I was not able to find the root cause of - when using the ui8 input, the histogram shows excessive quantization and there is a loss of shadow detail:
example_ui8
But when putting those exact same 8-bit values in a ui16 buffer, everything looks fine:
example_ui16

### DIFF
--- a/src/pipe/modules/i-vid/conv.comp
+++ b/src/pipe/modules/i-vid/conv.comp
@@ -26,44 +26,55 @@ main()
   float v = texture(img_in[2], (ipos+0.5)/vec2(imageSize(img_out))).r;
 
   vec3 rgb = vec3(0.0);
-  if(params.color == 1)
+  if(params.color == 0 || params.color == 1)
   { // says the ITU-R BT.2087-0
     vec3 YCbCr = vec3(Y,u,v);
-    // D'YCbCr -> E'YCbCr : de-scale range Q_YC
-    //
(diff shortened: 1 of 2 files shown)


hanatos:
looking good! about the ui8 issue, i think that was my mistake and should be fixed with https://github.com/hanatos/vkdt/commit/bfb650ef58bcd0b82609343fa6176e529ec8b5da . this should take care only the input is 8 bit, not the output after unapplying the TRC.

maybe we can use your version without the all-16-bit change  then? i'd like to minimise data shuffling from cpu to gpu.

paolodepetrillo:
That was it, I completely missed that line when trying to troubleshoot.

PR #94, opened 2023-10-28
Fixes for i-vid inputs (8 or 16 bit input)

paolodepetrillo:
Same as #93 but without the unnecessary change to always use a 16 bit input to the shader.


hanatos:
great, thanks for updating this again! merged.

PR #97, opened 2023-11-01
RFC / WIP - Lens correction

paolodepetrillo:
I have been experimenting with a lens correction module which uses a LUT to map a radius in the output image to a radius in the input image, rather than any specific distortion model, to correct for distortion and / or TCA.

This first attempt only supports the WarpRectilinear opcode embedded in DNG file metadata - I figured this was a good starting point since it's [openly documented](https://helpx.adobe.com/content/dam/help/en/photoshop/pdf/DNG_Spec_1_7_0_0.pdf) rather than being reverse engineered from proprietary tags. It could also easily be extended to support lensfun models, probably without any change to the shader kernel.

I have mostly been testing this with Olympus .orf's converted through Adobe DNG Converter - it reads the proprietary embedded correction and converts to WarpRectilinear.

There is a slight difference from the behavior of the darktable implementation: the correction results in a non-rectangular image with irregular black edges, and darktable automatically finds the largest rectangle inscribed in that image and sets the scale to crop to that rectangle and hide those black edges. This may be needed for some other models like lensfun, but DNG WarpRectilinear typically has the correct scaling implicit in the coefficients already so I don't try to do any auto scaling. The scale slider is provided to override if necessary.

Possible TODOs:

- Support rawloader in addition to rawspeed+exiv2 (Would need to modify rawloader to make the necessary exif tags available)
- Lensfun models
- Correction for Sony and Fuji raws [as in darktable](https://github.com/darktable-org/darktable/blob/master/src/iop/lens.cc) and Olympus [as in this darktable pr](https://github.com/darktable-org/darktable/pull/12760)
- Apply tangential coefficients in WarpRectilinear, although I've never seen a file where they were nonzero
- Support other DNG tags WarpRectilinear2 and WarpFisheye
- Option to apply only distortion, only TCA, or both (not possible for DNG but should be otherwise)

### DIFF
--- a/src/pipe/lens_corr.h
+++ b/src/pipe/lens_corr.h
@@ -1,0 +1,36 @@
+#pragma once
+
+typedef enum dt_lens_corr_type_t
+{
+  s_lens_corr_none = 0,
+  s_lens_corr_dng  = 1
+}
+dt_lens_corr_type_t;
+
+typedef struct dt_warp_rect_coef_t
+{
+  double kr0, kr1, kr2, kr3;
+  double kt0, kt1;
+}
+dt_warp_rect_coef_t;
+
+typedef struct dt_lens_corr_dng_t
+{
+  int                 warp_rect_enabled;
+  dt_warp_rect_coef_t warp_rect_coef[3];
+  double              warp_rect_center_x, warp_rect_center_y;
+}
+dt_lens_corr_dng_t;
+
+typedef union dt_lens_corr_params_t
+{
+  dt_lens_corr_dng_t dng;
+}
+dt_lens_corr_params_t;
+
+typedef struct dt_lens_corr_t
+{
+  dt_lens_corr_type_t   type;
+  dt_lens_corr_params_t params;
+}
+dt_lens_corr_t;
--- a/src/pipe/lens_corr.h
+++ b/src/pipe/lens_corr.h
@@ -16,7 +16,10 @@ dt_warp_rect_coef_t;
 
 typedef struct dt_lens_corr_dng_t
 {
-  int                 warp_rect_enabled;
+  // 0: no WarpRectilinear
+  // 1: warp_rect_coef[0] applies to all planes
+  // 3: separate coefficients for each plane
+  int                 warp_rect_planes;
   dt_warp_rect_coef_t warp_rect_coef[3];
   double              warp_rect_center_x, warp_rect_center_y;
 }
--- a/src/pipe/lens_corr.h
+++ b/src/pipe/lens_corr.h
@@ -16,6 +16,8 @@ dt_warp_rect_coef_t;
 
 typedef struct dt_lens_corr_dng_t
 {
+  int                 active_area[4]; // top, left, bottom, right
+
   // 0: no WarpRectilinear
   // 1: warp_rect_coef[0] applies to all planes
   // 3: separate coefficients for each plane
--- a/src/pipe/modules/lenswarp/main.c
+++ b/src/pipe/modules/lenswarp/main.c
@@ -1,0 +1,125 @@
+#include "modules/api.h"
+#include "core/log.h"
+#include "config.h"
+
+static void generate_radius_lut_dng(
+    dt_image_params_t *img_param,
+    float             *lut)
+{
+  dt_lens_corr_dng_t *dng = &img_param->lens_corr.params.dng;
+
+  if(dng->warp_rect_coef[0].kt0 != 0 || dng->warp_rect_coef[1].kt1 != 0
+    || dng->warp_rect_coef[0].kt0 != 0 || dng->warp_rect_coef[1].kt1 != 0
+    || dng->warp_rect_coef[0].kt0 != 0 || dng->warp_rect_coef[1].kt1 != 0)
+  {
+    dt_log(s_log_err, "[lenswarp] Ignoring WarpRectilinear tangential coefficients");
+  }
+
+  // For WarpRectilinear, radius 1 is defined as the distance from the center to
+  // the furthest corner of the uncropped image (or ActiveArea?). For the LUT we
+  // want to define radius 1 as the corner of the crop area.
+
+  // pixel location of center in the uncropped image
+  float cx = dng->warp_rect_center_x * img_param->uncropped[0];
+  float cy = dng->warp_rect_center_y * img_param->uncropped[1];
+  // distance to furthest corner of uncropped image
+  float dng_dist = hypotf(
+    MAX(cx, img_param->uncropped[0] - cx),
+    MAX(cy, img_param->uncropped[1] - cy));
+  // distance to furthest corner of cropped image
+  float crop_dist = hypotf(
+    MAX(cx - img_param->crop_aabb[0], img_param->crop_aabb[2] - cx),
+    MAX(cy - img_param->crop_aabb[1], img_param->crop_aabb[3] - cy));
+  float r_scale = crop_dist / dng_dist;
+
+  for(int ch = 0; ch < 3; ch++)
+  {
+    dt_warp_rect_coef_t *c = &dng->warp_rect_coef[ch];
+    for(int i = 0; i < LUT_SIZE; i++)
+    {
+      float r = (float)i / (LUT_SIZE - 1); // radius in the undistorted output image
+      float rs2 = r * r * r_scale;
+      lut[i * 4 + ch] = c->kr0 + rs2 * (c->kr1 + rs2 * (c->kr2 + rs2 * c->kr3));
+    }
+  }
+}
+
+static void get_center_dng(
+    dt_image_params_t *img_param,
+    float             *center)
+{
+  // TODO this might actually be relative to the ActiveArea not the full uncropped area?
+
+  dt_lens_corr_dng_t *dng = &img_param->lens_corr.params.dng;
+  // pixel location of center in the uncropped image
+  float cx = dng->warp_rect_center_x * img_param->uncropped[0];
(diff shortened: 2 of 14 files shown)


hanatos:
wow thanks for your work on this! it looks really full featured already. will check it out and comment in more detail after that.

yeah the rawloader codepath is kinda important, but it's not hard to grab the metadata from there too.

if you point exiv2 at the .lcp shipped with the adobe dng converter, would it give you the same values? i suppose we could also think about including this kind of data in the [camconst repo](https://github.com/hanatos/vkdt-camconst).

hanatos:
okay, a few general remarks (will put more detailed ones interleaved in the code):

very cool how you found your way around the code/api even with uploading custom source inside a module without exposing it to the outside!

the idea to store a few precomputed positions in a lut is very much a cpu idea.. i will bet that these handful of multiply-adds in the horner scheme will completely be hidden in the latency to get to the pixels on a gpu. it also simplifies the code and removes the need to upload anything at all.

about the coefficients/parameters: i would like it better if the coefficients were directly passed as parameters to the compute kernel, and maybe even exposed like this in the gui (possible when not using a lut). that way people can at least dial in the right coefficients even if they don't have an adobe raw converter. it may be easier to ask "a friend" with a windows computer to read out the values from the `.lcp` file (these are plain human readable xmp). it will also enable copy/pasting histories with correction parameters.

long term maybe we can add an option to read `.lcp` files or text files with the same content, for cameras which don't write the set of required dng tags.

are there other relevant correction models that require a very different set of arguments? i suppose exposing an int to distinguish between model and a float array with sufficient space for the most complex model would always be possible.

src/pipe/lens_corr.h
```diff
@@ -0,0 +1,41 @@
...
+  // 0: no WarpRectilinear
+  // 1: warp_rect_coef[0] applies to all planes
+  // 3: separate coefficients for each plane
+  int                 warp_rect_planes;
+  dt_warp_rect_coef_t warp_rect_coef[3];
+  double              warp_rect_center_x, warp_rect_center_y;
+}
+dt_lens_corr_dng_t;
+
+typedef union dt_lens_corr_params_t
```
hanatos, line 30: ah, here's my union already. i guess the parameters of the lenswarp module could look the same. not sure about double precision. looking at the coefficients in some random lcp it looks 32-bit float might suffice. doing double arithmetic on gpu is not much fun.

src/pipe/modules/lenswarp/main.c
```diff
@@ -0,0 +1,151 @@
...
+  float r_scale = crop_dist / dng_dist;
+
+  for(int ch = 0; ch < 3; ch++)
+  {
+    dt_warp_rect_coef_t *c = &dng->warp_rect_coef[ch];
+    for(int i = 0; i < LUT_SIZE; i++)
+    {
+      float r = (float)i / (LUT_SIZE - 1); // radius in the undistorted output image
+      float rs2 = r * r * r_scale;
+      lut[i * 4 + ch] = c->kr0 + rs2 * (c->kr1 + rs2 * (c->kr2 + rs2 * c->kr3));
```
hanatos, line 42: yeah see general comment. this is not nearly enough to make a gpu sweat per pixel. also since it appears only even powers of r are needed, maybe the square root above (in the hypot) is not actually required to be computed at all, so the special function unit would be idle too.

src/pipe/modules/lenswarp/main.c
```diff
@@ -0,0 +1,151 @@
...
+    MAX(cy - img_param->crop_aabb[1], img_param->crop_aabb[3] - cy));
+  float r_scale = crop_dist / dng_dist;
+
+  for(int ch = 0; ch < 3; ch++)
+  {
+    dt_warp_rect_coef_t *c = &dng->warp_rect_coef[ch];
+    for(int i = 0; i < LUT_SIZE; i++)
+    {
+      float r = (float)i / (LUT_SIZE - 1); // radius in the undistorted output image
+      float rs2 = r * r * r_scale;
```
hanatos, line 41: you sure this is correct? seems to me r*r is square and r_scale is still linear (because of the sqrt above?).

src/pipe/modules/lenswarp/main.c
```diff
@@ -0,0 +1,151 @@
...
+        pa->kr3 != pb->kr3 || pa->kt0 != pb->kt0 || pa->kt1 != pa->kt1)
+    {
+      // something is different, apply each set of coefs separately
+      return 3;
+    }
+  }
+  return 1;
+}
+
+int read_source(
```
hanatos, line 89: see general comment: passing the parameters straight can probably do completely without `read_source` and `init` and would only use `commit_params` to copy over stuff from the `img_param` struct to the `committed_param` uniforms of the kernel.

src/pipe/modules/lenswarp/main.c
```diff
@@ -0,0 +1,151 @@
...
+{
+  dt_roi_t lutroi = (dt_roi_t){ .wd = LUT_SIZE, .ht = 1 };
+  const int id_radlut = dt_node_add(graph, module, "radlut", "source",
+    LUT_SIZE, 1, 1, 0, 0, 1,
+    "source", "source", "rgba", "f32", &lutroi);
+
+  const int id_main = dt_node_add(graph, module, "lenswarp", "main",
+    module->connector[0].roi.wd, module->connector[0].roi.ht, 1,
+    0, 0, 3,
+    "input",  "read",  "rgba", "f16", &module->connector[0].roi,
```
hanatos, line 144: minor convenience thing: you can use `-1ul` (note the 64 bits) instead of `&module->connector[0].roi` to skip passing a roi for an input, it will just use the one of the connected counterpart later on.

src/pipe/modules/lenswarp/readme.md
```diff
@@ -0,0 +1,22 @@
+# lenswarp: correct for lens distortion and TCA
```
hanatos, line 1: yay you even have documentation :D

src/pipe/modules/lenswarp/main.comp
```diff
@@ -0,0 +1,77 @@
...
+layout( // radius LUT
+  set = 1, binding = 2
+) uniform sampler2D radlut_in;
+
+void
+main()
+{
+  ivec2 ipos = ivec2(gl_GlobalInvocationID);
+  if(any(greaterThanEqual(ipos, imageSize(img_out)))) return;
+  if(params.planes == 0)
```
hanatos, line 33: for maximum performance in this case, there is [a bypass api call](https://github.com/hanatos/vkdt/blob/master/src/pipe/modules/resize/main.c#L24) like used in the resize module in the link. it will cpu side short-circuit the connectors so the shader will not even be called, avoiding a roundtrip from and to global memory. has the downside that we'll need to check if parameters changed  and in this case recreate the pipeline, but shaves off 0.1ms or so..

src/pipe/lens_corr.h
paolodepetrillo: I only used a double here because that's how it was encoded in the DNG opcode.

src/pipe/modules/lenswarp/main.c
paolodepetrillo: I think that's a typo.

paolodepetrillo:
Thanks for the detailed feedback! I had already almost finished Olympus support so I will push that first and then start addressing the comments.

I had not thought about using the lcp files from Adobe DNG Converter. For Micro Four Thirds (and maybe also Fuji and others) there are no lcp files and only the coefficients supplied by the camera body in the exif are used. For Olympus I think the camera/lens has a 2D LUT of parameters indexed by focus distance and focal length, and for each individual shot it interpolates based on the exact settings for that shot.

There are a few different models that could potentially be supported:

- **[DNG](https://helpx.adobe.com/content/dam/help/en/photoshop/pdf/DNG_Spec_1_7_0_0.pdf) WarpRectilinear** - polynomial model with radial and tangential parts. Here I implemented the radial part and ignored the tangential part. Defines the corner of the raw image as radius 1.
- **[Adobe Camera Model](https://download.macromedia.com/pub/labs/lensprofile_creator/lensprofile_creator_cameramodel.pdf) (rectilinear)** - used in lcp files, appears to be the same model as WarpRectilinear but the radius is defined differently in terms of focal length independent of sensor size. Lensfun supports it so could follow the example of its implementation.
- **Olympus proprietary exif tag** - same polynomial as radial part of WarpRectilinear but radius is scaled slightly differently ([details here](https://github.com/darktable-org/darktable/pull/12760))
- **[Sony](https://github.com/darktable-org/darktable/blob/7d0a75c9704cd7635d8639c31456bc23a1113b6e/src/common/exif.cc#L847) or [Fuji](https://github.com/darktable-org/darktable/blob/7d0a75c9704cd7635d8639c31456bc23a1113b6e/src/common/exif.cc#L871) proprietary** - reverse engineered in darktable, 16 or 9 element LUTs in the exif tag
- **[Lensfun](https://lensfun.github.io/manual/latest/group__Lens.html#gaa505e04666a189274ba66316697e308e)** - poly3, poly5, ptlens models used in the lensfun database
- Latest DNG version also has a new WarpRectilinear2 tag with a different polynomial, more coefficients. There is also a WarpFisheye

It makes sense to have all of the coefficients be module parameters so that they can be set manually by editing the config. Probably the model with the largest number of parameters would be the 16 element Sony LUT. In some cases the coefficients could be modified in commit_params, for example to handle the different radius scaling between WarpRectilinear or ACM.

paperdigits:
I was just wondering if there was anymore work on this @paolodepetrillo? Looking at switching raw processors and lens correction is really needed badly for some of my cameras.

PR #100, opened 2023-12-17
Fix for upstream change in rawler

paolodepetrillo:
A recent change to rawler to support monochrome dng broke compatibility with lib.rs. This is barely tested but seems to fix it.


paolodepetrillo:
Oops never mind, didn't see it was already fixed!

PR #114, opened 2024-02-04
Correct way to determine stats on remote filesystems 

dydyamotya:
I have remote file storage mounted with nfs. 

For some files dirent reported DT_UNKNOWN type.

As mentioned here [stackoverflow article](https://stackoverflow.com/questions/47078417/readdir-returning-dirent-with-d-type-dt-unknown-for-directories-and) we need to call (l)stat manually for that systems.

I've added this ability. Please check if it is okey.

P.S. I'm not a pro in C, so if you know how to rewrite good - pls rewrite.

### DIFF
--- a/src/core/fs.h
+++ b/src/core/fs.h
@@ -280,7 +280,19 @@ static inline int fs_islnk(const char *dirname, const struct dirent *e)
 #warning "port me!"
   return 0;
 #else
-  return e->d_type == DT_LNK;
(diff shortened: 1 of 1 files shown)


hanatos:
hmm thanks for finding this. apparently this is valid behaviour and your patch is required. i'd probably replace the strcat stuff by something like snprintf such that we are sure to never exceed the buffer.

PR #121, opened 2024-02-23
Create colour-draw-guided.pst

blitzgneisserin:
This is a preset for modifying the color of certain areas that you paint over. Change the sliders in the module colour 02, e.g. saturation or white balance, and draw over the area.


hanatos:
great, thanks! may i suggest to name the instances of the involved modules to a commen name, something speaking such as `tint` or `draw` etc? this way you can easily find the lot by searching for instance name in the ui.

blitzgneisserin:
You mean like [this](https://github.com/hanatos/vkdt/pull/121/commits/c421a0c0a9626e8786492a648076e9dd72db3306)?

PR #123, opened 2024-03-03
Create draw-double-tint.pst

blitzgneisserin:
This is a preset that allows to change the color of specific areas of a photo in two different ways with the help of drawn masks/selections. Change e.g. saturation in colour tint1 and draw over an area, and change e.g. blue in colour tint2 and draw over an area.


hanatos:
..i was thinking that this should probably be more generic, so you can add the preset as many times as you want to add more of the same with different instance names. but i think that'd require some programming, will keep it for future use.

until then it's actually fun to work with this preset:
2024-03-14-193321_1920x1080_scrot

blitzgneisserin:
Screenshot_20240315_092915
Screenshot_20240315_093816

Originally I intended this preset more for corrective edits, e.g. when you pull up saturation you might get different color casts in different areas. E.g. here the orange sky requires a slightly different fix than the blue hill with the church in the background, plus the green foreground was too saturated compared to the middle ground.
But I guess it can be used for creative edits as well.

PR #130, opened 2024-06-16
vkdt-read-icc modifications for wlroots

blitzgneisserin:
Well, don't think that this was easy. Probably it could be much shorter, but I don't know how.

This works on wlroots-based compositors, but not on KDE Plasma 6, there I get this strange number, but no idea from where:

```
[anna@archlinux Downloads]$ ./vkdt-rawler-wayland-0.8.99-35-gb7ce8d37-x86_64.AppImage
[gui] glfwGetVersionString() : 3.3.6 Wayland EGL OSMesa clock_gettime evdev shared
[gui] monitor [0] BNQ BenQ SW240/21573 at 0 0
[gui] vk extension required by GLFW:
[gui]   VK_KHR_surface
[gui]   VK_KHR_wayland_surface
[gui] no joysticks found
[gui] no display profile file display.BNQ BenQ SW240/21573, using sRGB!
[ERR] could not open directory '(null)'!
```

### DIFF
--- a/bin/vkdt-read-icc
+++ b/bin/vkdt-read-icc
@@ -102,33 +102,110 @@ out.write('%f %f %f\n'%(O[1,0], O[1,1], O[1,2]))
 out.write('%f %f %f\n'%(O[2,0], O[2,1], O[2,2]))
 out.close()
 
-# check if the operating system is not Windows
-if platform.system() != "Windows":
-    # find monitor connection name with xrandr
-    proc = subprocess.Popen(['xrandr'], stdout=subprocess.PIPE)
-    out, err = proc.communicate()
-    lines = out.decode().split('\n')
-    monitor_names = [line.split(' ')[0] for line in lines if ' connected' in line]
+def get_monitors_info():
+    try:
+        # Execute the command hwinfo --monitor and capture the output
+        output = subprocess.check_output(["hwinfo", "--monitor"]).decode("utf-8")
 
-    # ask user for which monitor the profile should be installed
-    print("available monitors:")
-    for i, monitor_name in enumerate(monitor_names):
-        print(f"{i+1}. {monitor_name}")
-    choice = 0
-    if len(monitor_names) > 1:
-      choice = int(input("for which monitor do you want to install the profile? ")) - 1
-    monitor_name = monitor_names[choice]
+        # Split the output into lines
+        lines = output.splitlines()
 
-    # path to directory ./config/vkdt in the user directory
-    config_dir = os.path.expanduser('~/.config/vkdt')
+        # Initialize lists to store monitors
+        vendor_lines = [line for line in lines if "Vendor:" in line]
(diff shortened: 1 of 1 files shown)


hanatos:
oh, nice, thanks for putting this together. indeed looks like this is substantially more complicated to do than the old version was. need to look at it in detail. wlroots should be good enough for everyone of course.. but i suppose when writing so much code for exhaustive functionality it'd be good if it worked on kde too.

hanatos:
just noticed that i don't even have hwinfo installed. this might be true for more people who occasionally use the script. maybe, in the interest of maximum compatibility, should we write a small c test program using *exactly* the same mechanism to find the monitor name as the main vkdt program would?

i.e. use some minimal `glfwInit()` and `glfwGetMonitorName(monitors[0]);` without even creating a window? not sure what this would do on macintosh or windows computers, but probably would be better than now.

needs c code to extract the icc tags but these files are simple enough.

blitzgneisserin:
> a small c test program using _exactly_ the same mechanism to find the monitor name as the main vkdt program

I had a similar idea but didn't know how to do that.

I just wrote a small test script in Python to find out the monitor name with glfw.get_monitor_name(monitor), but the script only found out the name of the connection, e.g. HDMI-A-1. But probably I am only being stupid and this needs to be done differently.

hanatos:
i think that's exactly the string. i know nothing about python, but these
HDMI-0 (nvidia) or HDMI-A-1 (amd?) identifiers look familiar.

blitzgneisserin:
I am not sure I understand what you mean. The connection names do not work for the new native Wayland appimages. They see the monitor as e.g. `[gui] monitor [0] BNQ BenQ SW240/21573 at 0 0`, not as HDMI-A-1

blitzgneisserin:
I just did find a way to get the exact monitor name on KDE Plasma, with Python, but some Python modules/packages are needed for that.

```
import gi
gi.require_version('Gdk', '3.0')
from gi.repository import Gdk

def get_monitor_names():
    display = Gdk.Display.get_default()
    monitor_names = []

    for i in range(display.get_n_monitors()):
        monitor = display.get_monitor(i)
        monitor_name = monitor.get_model()
        monitor_names.append(monitor_name)

    return monitor_names

def write_monitor_names_to_file(filename):
    monitor_names = get_monitor_names()
    with open(filename, 'w') as file:
        for name in monitor_names:
            file.write(name + '\n')

if __name__ == '__main__':
    write_monitor_names_to_file('monitornames.txt')
```

blitzgneisserin:
Oh man, now the problem is that the file name contains a /.

hanatos:
closing this for now because `read-icc` is now a c application (so i don't have to put python+numpy+iccdump as dependency in the arch linux aur package). sounds to me like the screen ids depend on compositor, but having this in c would allow us to grab them from the same source. i suppose it's also simple enough to just copy the file to the appropriate name once when doing the setup.

PR #156, opened 2024-11-20
fix some macos runtime and compile time warnings

marrony:
fixes compile time error:
```
gui/nuklear_glfw_vulkan.h:1264:52: error: use of undeclared identifier 'GLFW_MODS_CONTROL'
 1264 |   if (button == GLFW_MOUSE_BUTTON_LEFT && ((mods & GLFW_MODS_CONTROL) == 0))
```
also, removes some compile time warnings (not all).

removes runtime warnings about exceeding sampler binding count like that:
```
[qvk] validation layer: Validation Error: [ VUID-VkPipelineLayoutCreateInfo-descriptorType-03016 ] | MessageID = 0x56816e61 | vkCreatePipelineLayout():  max per-stage sampler bindings count (33) exceeds device maxPerStageDescriptorSamplers limit (16).
The Vulkan spec states: The total number of descriptors in descriptor set layouts created without the VK_DESCRIPTOR_SET_LAYOUT_CREATE_UPDATE_AFTER_BIND_POOL_BIT bit set with a descriptorType of VK_DESCRIPTOR_TYPE_SAMPLER and VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER accessible to any given shader stage across all elements of pSetLayouts must be less than or equal to VkPhysicalDeviceLimits::maxPerStageDescriptorSamplers (https://vulkan.lunarg.com/doc/view/1.3.296.0/mac/1.3-extensions/vkspec.html#VUID-VkPipelineLayoutCreateInfo-descriptorType-03016)
```

removes runtime warnings about freeing memory with buffers bound:
```
CHASSIS(WARN / SPEC): msgNum: 1936111581 - Validation Warning: [ CHASSIS ] Object 0: handle = 0xab64de0000000020, type = VK_OBJECT_TYPE_BUFFER; Object 1: handle = 0xc4f3070000000021, type = VK_OBJECT_TYPE_DEVICE_MEMORY; | MessageID = 0x7366b7dd | vkFreeMemory():  VK Object VkBuffer 0xab64de0000000020[] still has a reference to mem obj VkDeviceMemory 0xc4f3070000000021[].
    Objects: 2
        [0] 0xab64de0000000020, type: 9, name: NULL
        [1] 0xc4f3070000000021, type: 8, name: NULL
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.a
 *ctooltips
 *ptooltips
+*.dSYM/
 cscope.out
 built/
 src/core/version.h
--- a/bin/config.mk.defaults.osx
+++ b/bin/config.mk.defaults.osx
@@ -68,8 +68,8 @@ export VKDT_USE_MCRAW
 FFMPEG_VERSION=$(shell ffmpeg -version | head -1 | cut -f 3 -d' ' | cut -d. -f1)
 VKDT_USE_FFMPEG=$(eval VKDT_USE_FFMPEG := $$(shell [ $(FFMPEG_VERSION) -gt 5 ] && echo 1 || echo 0))
 export VKDT_USE_FFMPEG
-VKDT_AV_CFLAGS=$(eval VKDT_AV_CFLAGS := $$(shell pkg-config --cflags libavformat --cflags libavcodec --cflags libswresample))$(VKDT_AV_CFLAGS)
-VKDT_AV_LDFLAGS=$(eval VKDT_AV_LDFLAGS := $$(shell pkg-config --libs libavformat --libs libavcodec --libs libswresample))$(VKDT_AV_LDFLAGS)
+VKDT_AV_CFLAGS=$(eval VKDT_AV_CFLAGS := $$(shell pkg-config --cflags libavformat --cflags libavcodec --cflags libavutil --cflags libswresample))$(VKDT_AV_CFLAGS)
+VKDT_AV_LDFLAGS=$(eval VKDT_AV_LDFLAGS := $$(shell pkg-config --libs libavformat --libs libavcodec --libs libavutil --libs libswresample))$(VKDT_AV_LDFLAGS)
 export VKDT_AV_CFLAGS VKDT_AV_LDFLAGS
 
 # where to find glfw for the gui:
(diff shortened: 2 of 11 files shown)


hanatos:
nice, thanks! let me test this locally on a few machines.

src/gui/gui.c
```diff
@@ -158,7 +158,7 @@ dt_gui_win_init_vk(dt_gui_win_t *win)
   };
   VkDescriptorPoolCreateInfo pool_info = {
     .sType         = VK_STRUCTURE_TYPE_DESCRIPTOR_POOL_CREATE_INFO,
-    .flags         = VK_DESCRIPTOR_POOL_CREATE_FREE_DESCRIPTOR_SET_BIT,
+    .flags         = VK_DESCRIPTOR_POOL_CREATE_FREE_DESCRIPTOR_SET_BIT | VK_DESCRIPTOR_POOL_CREATE_UPDATE_AFTER_BIND_BIT,
```
hanatos, line 161: why did you have to change this one? trying a few things locally with your patch, this causes validation layer complaints on my end. this appears to become a problem  when uploading individual elements for array textures, where the command buffer is used in graph-run-nodes-upload.h:84. even `vkResetCommandBuffer` will not work without this flag so i'm not sure this is a good idea. i suppose any kind of animations might trigger an error.

marrony: Okay, I may have changed too much. This change was a consequence of this warning:

```
BestPractices-vkCreateCommandPool-command-buffer-reset(WARN / PERF): msgNum: 141128897 - Validation Performance Warning: [ BestPractices-vkCreateCommandPool-command-buffer-reset ] | MessageID = 0x86974c1 | vkCreateCommandPool(): pCreateInfo->flags has VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT set. Consider resetting entire pool instead.
```

I removed the `VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT` flag from command pool and it lead to this errors while calling `vkBeginCommandBuffer`, I poke around many flags and I fixed it calling `vkResetCommandPool` on graph.c:371 and forgot to revert this change.

src/pipe/graph.c
```diff
@@ -368,6 +368,8 @@ VkResult dt_graph_run(
       return VK_INCOMPLETE;
     }
 
+    QVKR(vkResetCommandPool(qvk.device, graph->command_pool, 0));
```
hanatos, line 371: okay maybe your validation layers are newer than mine or i didn't enable the best practices notices there.. anyways, if you want to reset the whole pool, we'll need one pool per command buffer. the graph has two for the odd/even pipelines when interleaving two subsequent frames on the gpu at the same time. especially amd seems to profit a lot from this scheme. in any case we'll need two command pools, one for each of the two command buffers, and then reset the one for the command buffer in question. probably there are some more places in the graph where commands are issued (uploading array images).

marrony: my intention with this PR is not introduce a big change like this, I will revert all the command pool changes and leave only the sampler bindings warning fixes, I can get back to to this in another moment.

PR #178, opened 2025-03-16
Do not check for queue transfer bit

paolodepetrillo:
Do not check for VK_QUEUE_TRANSFER_BIT which might not be set.

According to https://registry.khronos.org/vulkan/specs/latest/man/html/VkQueueFlagBits.html -
All commands that are allowed on a queue that supports transfer operations are also allowed on a queue that supports either graphics or compute operations. Thus, if the capabilities of a queue family include VK_QUEUE_GRAPHICS_BIT or VK_QUEUE_COMPUTE_BIT, then reporting the VK_QUEUE_TRANSFER_BIT capability separately for that queue family is optional.

Also add -lm to ldflags for a few modules which use math.h.


hanatos:
makes sense to me!

PR #181, opened 2025-03-22
Portability fixes for non-glibc Linux systems

paolodepetrillo:
The main purpose of this change is to allow vkdt to run on Android with bionic libc, but it also allows it to compile and run on a [musl](https://www.musl-libc.org/)-based distro like Void Linux. Although it often crashes with a stack overflow because [musl has small thread stacks by default](https://wiki.musl-libc.org/functional-differences-from-glibc.html#Thread-stack-size).

The portable qsort_r is only needed for bionic - it was recently added to bionic but that won't be available until Android 16 is released. It was copied from https://github.com/noporpoise/sort_r.

bionic doesn't have dlinfo at all, and musl has dlinfo but without RTLD_DI_ORIGIN.

There is probably a more elegant way to deal with the POSIX realpath without the GNU extension, but as far as I can tell this seems to work.

### DIFF
--- a/src/core/fs.h
+++ b/src/core/fs.h
@@ -184,6 +184,7 @@ fs_basedir(
   fs_dirname(basedir);
   char mod[PATH_MAX];
   snprintf(mod, sizeof(mod), "%s/darkroom.ui", basedir);
+#ifdef __GLIBC__
   if(access(mod, F_OK))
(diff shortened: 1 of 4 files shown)


hanatos:
thanks, looks good to me!

yeah i'm not really careful about stack usage. i don't really understand why i would want a harsh limit here. i suppose it's easy but very tedious to fix for musl.

Issue #167, opened 2025-03-03
Add Nix flake


WillPower3309:
I noticed from the readme that it appears vkdt is supportive of the nix project. Is there interest in including a nix flake in this project? This would involve a flake.nix and flake.lock file living in the project that would define and lock dependencies of the project so that nix users could build (and develop) the current main git branch, as opposed to having to use the most recent nixpkgs release. This would also have some [CI advantages](https://determinate.systems/posts/nix-github-actions/), if the project ever adds some more GitHub actions, which I'd also be happy to lend a hand with if there's interest.

I'd be more than happy to contribute and maintain the flake, but wanted to do an interest check before I cut a PR.

hanatos:
hi! i'm certainly interested in better access for nix (and arch) people. this appimage business breaks more often than i'd like, and yeah.. the release is super old now and lacks all the interesting features.

paperdigits:
@hanatos this is done and can be closed ;)

PR #183, opened 2025-03-30
add nix devShell

WillPower3309:
FIrst part of #167, this PR adds a minimal nix flake that allows nix users to enter a shell with all of this project's dependencies installed (pulled from the current package in nixpkgs). This allows for easy develop. You can now compile vkdt on a bare nix install by simply running `nix develop` and then `cd bin; make`. This provides a nice isolated dev environment for both linux and macOS nix users, running on both x86 and arm:
```
❯ nix flake show
git+file:///home/will/Projects/vkdt?ref=refs/heads/master&rev=c4c3936fc5dcdac4f4459e1d2094bcc961209e46
└───devShell
    ├───aarch64-darwin omitted (use '--all-systems' to show)
    ├───aarch64-linux omitted (use '--all-systems' to show)
    ├───x86_64-darwin omitted (use '--all-systems' to show)
    └───x86_64-linux: development environment 'nix-shell'
```

A follow-up would be to allow nix users to install the package through the flake as well, to ensure they get access to the bleeding edge version of vkdt. From here, we could also [incorporate basic compilation / syntax checks that PRs would need to pass prior to approval through a nix environment managed by the flake](https://determinate.systems/posts/nix-github-actions/) (if desired).

Caveats:
- If any dependencies are added to vkdt, we will need to include them in the flake until a new release is cut, as the flake just copies the dependencies from the currently released vkdt version found in the nix package repo. Considering the suckless goal of this project I figure this won't be a huge issue, but wanted to bring awareness to it. I would be more than happy to update the flake upon any extra dependencies being added
- Flakes rely on a lock file to lock down the versions of the flake's dependencies. I can include a github action to update this lock file periodically (bi weekly, monthly?) to reduce maintenance burden

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,20 @@
+{
+  description = "raw photography workflow that sucks less";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, utils, ... }:
+    utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = import nixpkgs { inherit system; };
+      in
+      {
+        devShell = with pkgs; mkShell {
+          inputsFrom = [ pkgs.vkdt ];
+        };
+      }
+    );
+}
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1743095683,
+        "narHash": "sha256-gWd4urRoLRe8GLVC/3rYRae1h+xfQzt09xOfb0PaHSk=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "5e5402ecbcb27af32284d4a62553c019a3a49ea6",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "nixpkgs": "nixpkgs",
+        "utils": "utils"
+      }
+    },
+    "systems": {
+      "locked": {
+        "lastModified": 1681028828,
+        "narHash": "sha256-Vy1rq5AaRuLzOxct8nz4T6wlgyUR7zLU309k9mBC768=",
+        "owner": "nix-systems",
+        "repo": "default",
+        "rev": "da67096a3b9bf56a91d16901293e51ba5b49a27e",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-systems",
+        "repo": "default",
+        "type": "github"
+      }
+    },
+    "utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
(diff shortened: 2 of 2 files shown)


flake.nix
```diff
@@ -0,0 +1,20 @@
...
+    utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, utils, ... }:
+    utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = import nixpkgs { inherit system; };
+      in
+      {
+        devShell = with pkgs; mkShell {
```
paperdigits, line 15: You should use `devshells.default` here.

flake.nix
```diff
@@ -0,0 +1,20 @@
...
+          inputsFrom = [ pkgs.vkdt ];
```
paperdigits, line 16: If this flake is to target the development branches of vkdt, we should really list out all of the dependencies of vkdt and not rely on the package from nixpkgs. the package in nixpkgs only gets updated when there is a release. This will also have the effect of making it easier to do the `nix run ...` part of the flake as well.

paperdigits (commented):
Hey, thanks for proposing this. I'm the maintainer of the vkdt package in nixpkgs. I think there are a few improvements that'll make this a much better flake. Thanks!

flake.nix
WillPower3309: fair enough

WillPower3309:
Added package output to the flake. @paperdigits `nix build` and `nix run` have an error:
```
❯ nix build
error: builder for '/nix/store/w0ngjm1mjhc67ak6zwv57m83r95b5ddv-vkdt-git.drv' failed with exit code 1;
       last 6 log lines:
       > Running phase: unpackPhase
       > unpacking source archive /nix/store/4apm98i85p5qlm71qzysiynlij4l3wsi-source
       > source root is source
       > Executing cargoSetupPostUnpackHook
       > unpacking source archive
       > do not know how to unpack source archive
       For full logs, run 'nix log /nix/store/w0ngjm1mjhc67ak6zwv57m83r95b5ddv-vkdt-git.drv'.
```

I believe this is because I don't include cargo build settings as we do in the nixpkgs definition of vkdt: https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/vk/vkdt/package.nix#L40-L46

Do you know how I can add that to the flake?

paperdigits:
@WillPower3309 why not just include the cargo build stuff?

paperdigits:
Also, you can rebase your changes on top of master, we probably don't want merge commits.

WillPower3309:
> Also, you can rebase your changes on top of master, we probably don't want merge commits.

Yup, planning a squash!

WillPower3309:
> @WillPower3309 why not just include the cargo build stuff?

Not sure how to get cargoDeps and cargoRoot options available in the flake, can't copy it directly and don't have any experience with packaging rust. Let me know if you know how to translate that bit to a flake

WillPower3309:
@paperdigits would really value a review from you if you can find some time :) finally all in a good working state!

hanatos:
awesome, nice progress!

for some context re:mcraw: this is to read raw video from the (commercial) MotionCam app. these folks do a great job at squeezing good/honest raw pixels out of telephone cameras. they provide a toolkit which i forked here: https://github.com/hanatos/motioncam-decoder

my fork essentially removes the decoding work from it and serves the encoded data to the GPU in vkdt. the actual decoding is done on device, which enables like 1000fps playback rates, provided fast enough hard drive.

one consequence however is that this code is probably less than useful for anyone outside vkdt (since it only contains a loader for encoded data, no decoder).

it's very much optional but of course would be nice to have.

in the PKGBUILD for arch, i had to remove stripping: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=vkdt-git because the c++ stuff tripped over it, not sure the issue here is the same. would be great to do without stripping.

Issue #186, opened 2025-04-14
filmsim: process for creating filmsim.lut


jedypod:
I saw that there are a few new film negative profiles in arctic's agx-emulsion so I thought I would try to create a new `filmsim.lut` file using the [instructions](https://jo.dreggn.org/vkdt/src/pipe/modules/filmsim/readme.html).

I modified the `film_stocks` list in `mklut-profiles.py` to add these 3
```
    'kodak_vision3_250d_uc',
    'kodak_vision3_200t_uc',
    'kodak_vision3_500t_uc',
```

After creating `~/.config/vkdt/data/filmsim.lut` the behavior of the filmsim module changes and I get washed out dark looking images :C

Is there anything I'm doing wrong? Any chance we could get the shipped filmsim.lut updated to include those 3?

Thanks for the help!

hanatos:
ah probably it's this line:

```
const int s_paper_offset = 12; // first paper in data list/lut
```
in `src/pipe/modules/filmsim/head.glsl`.

since i only load one lut with film stocks and paper in it, i need to side load which row in the texture will be the first paper. if you add three new films, this should be `15` instead.

also, for sanity you would likely change `params.ui` in the same folder to contain the film names in the combo box:
```
film:combo:kodak ektar 100:kodak portra 160:kodak portra 400:kodak portra 800:kodak portra 800 push 1:kodak portra 800 push 2:kodak gold 200:kodak ultramax 400:kodak vision3 50d:fujifilm pro 400h:fujifilm xtra 400:fujifilm c200
```
just append colon separated to the end, if you added the film to the end of the list in `mklut-profiles.py`. if you prefer to sort it together with the other vision3 it'll move up the others and invalidate people's old config files, which i've done before in this context and i'd say is fine before 1.0.

i'm a slight bit busy in the next few days still.. if you beat me to this feel free to make it a pull request! the last missing bit is probably update the incomplete instructions and create a new `src/filmsim.lut.xz` in git so it will be shipped and extracted during build for everyone.

hanatos:
thanks for preparing the PR. i updated the documentation to include the additional steps when updating.

PR #187, opened 2025-04-15
Update filmsim profiles to latest from agx-emulsion

jedypod:
This PR updates the filmsim negative profiles to the latest provided by [agx-emulsion](https://github.com/andreavolpato/agx-emulsion), adding
```
    'kodak_vision3_250d_uc',
    'kodak_vision3_200t_uc',
    'kodak_vision3_500t_uc',
```
I've chosen to sort them in order, so this PR will break existing cfg files.

After some digging, I also discovered that I had to update the contents of `wb.h` with the whitebalance information for the new profiles. The bash script took a while to run, and I chose to use the calculated values for all the film profiles. If you want me to only change the ones I added let me know, I can tweak that.

Seems to work okay in my testing! Please let me know if I missed anything!

Refs #186


hanatos:
whoa very nice! thanks for also running the white balance optimiser, very cool that you discovered it, i had forgotten this aspect.

oops, yes, no semicolon in python.

tested and works as intended. thanks heaps!

PR #200, opened 2025-10-13
Oklab curves

hanatos:
this pull request comes with a few breaking changes to the `curves` module. i *think* it's better now but am interested in feedback.

* the working space is now not YCbCr but oklab.
* it works on *saturation* not *chroma*, i.e. compensated for luminance, or lightness as in Lab really. this avoids desaturation when messing with the lightness channel.
* chroma is pre-procesed using a guided filter, to avoid colour noise amplifying itself.

note that most of this is really similar in dt's implementation. main differences: i'm using oklab because it's a bit simpler to work with. downside: it's like not hue constant at all. messing with Y or C based on h will likely change colour. i guess for smaller changes it's not much, and it can always be corrected via h/h or something.

no gamut mapping (yet?). pushing saturation will simply push values out the window and you'll lose definition. can introduce some hyperbolic mapping to force values back into sane range, but i'd see this as an extra feature that can likely be put on top and made optional via combo box or something.

i really want the `curves` module in `core`, i.e. with backward compatibility guarantees. formally only after the 1.0 release, but do let me know if you have YCbCr edits that you like to keep.

note on breaking changes: unlike dt, the parameters here are named and will stick/stay compatible if you switch between this branch and current master. just the computed result will look different, but there is no anticipated data loss when trying this branch.

if you have a minute, let me know your thoughts.

### DIFF
--- a/src/pipe/modules/curves/prep.comp
+++ b/src/pipe/modules/curves/prep.comp
@@ -1,0 +1,44 @@
+#version 460
+#extension GL_GOOGLE_include_directive    : enable
+#extension GL_EXT_nonuniform_qualifier    : enable
+#include "shared.glsl"
+layout(local_size_x = DT_LOCAL_SIZE_X, local_size_y = DT_LOCAL_SIZE_Y, local_size_z = 1) in;
+layout(std140, set = 0, binding = 1) uniform params_t
+{
+  vec4 abcdr[8];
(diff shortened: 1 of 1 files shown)


WillPower3309:
LGTM! I'll open a separate issue for the nix flake

PR #215, opened 2025-12-15
Fix #214

### LINKED ISSUE #214: window positioned in un-draggable location
Initial window position is placed into a location that makes it impossible to grab the window decorator.

Currently:
Image

Ideally:

Image

phcreery:
Tested on
- [x] Windows
- [x] Windows hidpi
- [ ] wayland
- [ ] x11
- [ ] Mac

### DIFF
--- a/src/gui/gui.c
+++ b/src/gui/gui.c
@@ -50,7 +50,7 @@ framebuffer_size_callback(GLFWwindow* w, int width, int height)
   if(w == vkdt.win1.window) win = &vkdt.win1;
   if(width != win->width || height != win->height)
   {
-    // win->width = width; win->height = height;
+    win->width = width; win->height = height;
     dt_gui_recreate_swapchain(win);
     nk_glfw3_resize(w, win->width, win->height);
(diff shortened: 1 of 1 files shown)


src/gui/main.c
```diff
@@ -338,7 +338,16 @@ int main(int argc, char *argv[])
       dt_gui_present();
 
     static int bs = 1;
-    if(bs) { dt_gui_toggle_fullscreen(); bs = 0; }
+    if (bs) {
```
hanatos: uhm i don't understand this part. isn't it just a verbatim copy of the toggle_fullscreen call that has been removed? up to setting the x and y coordinate too? if so, why not change the offsets in api.h:dt_gui_toggle_fullscreen directly?

hanatos:
okay tried on wayland.

it completely ignores the x and y position coordinates, as expected.

also, the size restoration now has some weird effect on the font size that appears to be double size on first initialisation, then resizing to about the same window size goes back to normal. looking..

PR #216, opened 2025-12-16
add ability to delete nodes with delete or backspace key

phcreery:


hanatos:
just for some extended context: in general it's better to use nuklear functions instead of glfw directly when it comes to events. nk provides stuff like "button has been pressed in this frame" which is a one-time event. your patch will trigger the branch a couple of times for consecutive frames, while the button is still held down. this is not a problem because you're testing whether there is anything selected and the first invocation will clear the selection. also i don't think nuklear knows much about backspace or delete keys, so i'm merging this, thanks!

hanatos:
nice. actually this is better, let me update.

PR #221, opened 2025-12-23
port remaining fs functions to windows

phcreery:
I wasn't able to test `fs_createdate` yet. Let me spend some more time and I will test it.

### DIFF
--- a/src/pipe/modules/o-jpg/main.c
+++ b/src/pipe/modules/o-jpg/main.c
@@ -197,8 +197,14 @@ void write_sink(
 
     // maybe route this string in via o-jpg params (beware of the ':' and then dt_sanitize_user_string + dt_strexpand it)
     char cmd[1024];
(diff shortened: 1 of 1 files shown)


phcreery:
Oh thanks, didn't see that one. `strptime` doesn't seem to be provided by the ucrt64 lib... need to think some more on this one.

hanatos:
good stuff here! let me know once you think this is ready to merge. the macintosh build is failing because it can't find `<vulkan/vulkan.h>`, i'm not sure this is related to any of your changes at all though.

phcreery:
What are your thoughts on removing the exiftool bin search and store in `gui/render_lighttable.c:1170`? I think the "trampoline" `#!/bin/sh` script in `./bin` is nicer. Especially since exiftool will be bundled on windows

phcreery:
Okay, I basically added the changes from https://github.com/hanatos/vkdt/pull/224. but also included quotes in case the exiftool binary path has spaces. This is ready now

hanatos:
ah good idea to use the trampoline for all platforms! will test and merge then..

PR #224, opened 2025-12-30
Fixed exiftool not copying data from paths with spaces

Eemilp:
Added necessary quoting to properly copy exif data when the paths contain spaces.


hanatos:
i believe this has been merged, closing the pr. thanks for putting this in!

Issue #225, opened 2026-01-02
Mouse lag


phcreery:
I notice a lag, or delay rather, when between my mouse cursor and gui items like dragging nodes, tooltip locations, sliders, etc.

PR #227, opened 2026-01-02
gui: add option for vk present mode

phcreery:
I don't know much about vulkan rendering, but I found that changing the present mode from VK_PRESENT_MODE_FIFO_KHR to VK_PRESENT_MODE_MAILBOX_KHR seems to reduce the input delay drastically.

I don't know if there is something deeper in the rendering pipeline that can help.

Resolves https://github.com/hanatos/vkdt/issues/225


hanatos:
looks good to me. might be a windows specific issue? i think on wayland it even depends on compositor. good to have a config option.

PR #229, opened 2026-01-04
qvk: rename config present_mode to vk_present_mode

phcreery:
`present_mode` was too generic. I think this is more fitting.


hanatos:
uhm okay, this is your feature here, so i won't argue about variable names. i would probably not care enough to spend 3 more uint8_t on this :)

PR #235, opened 2026-02-03
Update cargoDeps hash and flack.lock file.

paperdigits:
I've updated the cargoDeps hash, which is a hash of the dependencies for the raw decoder in that module. Nix wants a hash of everything so it can verify what it is building. I got the new hash from the error message while trying to build the old hash. You'll need to update this hash each time you update that raw decoder module.


hanatos:
yay, thanks!

PR #258, opened 2026-04-09
pipe: recycle graphs instead of resetting them

Randalix:
Follow-up to #254 — you noted that thumbnail rendering tears down graphs
frequently, so the `vkQueueWaitIdle` fix there could get expensive.
This addresses that.

`dt_graph_reset()` was doing a full cleanup + reinit on every image load:
destroying semaphores, command pools, query pools, and device memory, then
recreating them. That's unnecessary — these objects are stable across images.

This replaces it with `dt_graph_repurpose()` which only clears the
logical/per-image state (modules, nodes, pipelines, connector images) while
keeping the Vulkan handles alive. The allocator bookkeeping is reset but the
underlying GPU memory is kept, so same-sized images avoid the alloc/free cycle.

Also refactors the shared teardown logic in `dt_graph_cleanup()` into three
small helpers to avoid duplication between the two paths.

### DIFF
--- a/src/db/thumbnails.c
+++ b/src/db/thumbnails.c
@@ -267,7 +267,7 @@ dt_thumbnails_cache_one(
     if(tcfg && (tbc1 >= tcfg)) return VK_SUCCESS; // already up to date
   }
 
-  dt_graph_reset(graph);
+  dt_graph_repurpose(graph);
 
   char *extrap[] = {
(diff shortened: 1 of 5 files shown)


hanatos:
this is a delicate change but appears to work well, and is validation layer warning free :)

PR #261, opened 2026-04-09
fix dynamic lod for sliders and drag keys

Randalix:
fixes #260

two bugs. the dragkey feature left a per-frame set_lod call at the end of render_darkroom which was resetting lod every frame, so sliders would drop to interact-lod for one frame and immediately snap back. the other one is that interact_end restoring lod_fine in render_darkroom.h is inside a hover check, so releasing a slider outside its bounds left the preview stuck at interact-lod.

fix is an unconditional interact_end restore before widget rendering, and latch transition detection for dragkeys. four lines.

### DIFF
--- a/src/gui/render_darkroom.c
+++ b/src/gui/render_darkroom.c
@@ -416,6 +416,10 @@ void render_darkroom()
     return;
   }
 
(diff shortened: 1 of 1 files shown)


hanatos:
nice, this works here. thanks for all these super quick fixes!

PR #262, opened 2026-04-09
fix: chord menu should not grab modified hotkeys

Randalix:
fixes #259

when a modifier key is held, the chord menu now ignores the keypress instead of matching it as a leader key. so ctrl+p opens the preset picker instead of the 'p' submenu.


hanatos:
nice, thanks!

PR #264, opened 2026-04-24
    grade: add color wheels and zoned log mode

Randalix:
Replaces the lift/gamma/gain rgb knobs with proper color wheels and adds
a zoned log mode that keeps each wheel in its tonal lane.

The new `colwheel` widget draws a hue ring. Drag to push color, Shift+drag for faster sweeps, double-click to
reset. A master brightness slider sits below each wheel. Wheels pair up
two per row.

The grade module now has four wheels: lift, gamma, gain, and a new
offset for nudging the output after the main math — handy for black
and white point without touching gamma.

zoned log mode
In classic mode the wheels act globally, same as before. Zoned log maps
luminance to a perceptual scale and splits it at two adjustable pivots,
so lift only touches shadows, gamma only touches midtones, and gain only
touches highlights. Less bleed, more intuitive.


bin/default-darkroom.i-raw
```diff
@@ -20,12 +21,13 @@ connect:crop:01:output:colour:01:input
...
+connect:grade:01:output:display:main:input
+connect:grade:01:output:hist:01:input
 connect:hist:01:output:display:hist:input
 param:colour:01:exposure:0
 param:llap:01:sigma:0.12
 param:llap:01:shadows:1
 param:llap:01:hilights:1
 param:llap:01:clarity:0.2
-param:i-raw:main:filename:none.raw
+param:i-raw:main:filename:none.raw
```
hanatos: needs newline at the end of the file, else it won't parse :)

hanatos:
really cool!

the mouse doesn't grab, so the pointer is limited in movement a bit by the edge of the screen. i don't think i'd need the full range though for usual edits, so not much of a problem.

the spacing is a bit weird here:
space

src/gui/render_darkroom.h
```diff
@@ -459,6 +459,193 @@ render_darkroom_widget(int modid, int parid, int is_fav_menu)
...
+      int already_active = vkdt.wstate.active_widget_modid == modid &&
+                           vkdt.wstate.active_widget_parid == parid &&
+                           vkdt.wstate.active_widget_parnm == num;
+      // double-click resets to default (like other sliders)
+      if(!vkdt.wstate.dragkey_latched &&
+         nk_input_is_mouse_click_in_rect(&ctx->input, NK_BUTTON_DOUBLE, warea))
+      { val[0] = def[0]; val[1] = def[1]; val[2] = def[2]; val[3] = def[3]; change = 1;
+        if(already_active) widget_end(); }
+      // right-click contextual menu
+      { struct nk_vec2 csize = {warea.w, warea.w};
```
hanatos: you prefer this in line instead of using the shared macro? i mean we could probably make this a function call or something that is better readable. would be good to keep the code shared though, so we can consistently add/remove/rename entries in the context menu.

hanatos:
super minor: are these wheels supposed to have thin black lines between the colour swatches? or should the fill the space fully by joining at the edges? now it appears to be both, depending on aliasing.

Randalix:
Thanks for you feedback! All issues should be adressed now.

hanatos:
thanks for the improvements! another round of quick comments:

* the sliders and wheels should probably grab the cursor too if they are clicked (not only via drag key). i feel like i have to move a fair bit with the mouse to reach the edge of the ring. does that depend on pixel resolution? 4k here.
* if accessed via dragkey (can i do this with my gamepad stick :) ?) it seems that the mouse x and y coordinates are swapped? i can't get the white dot move in coherent ways then.
* the dragkey does not enforce limits. i can move the white dot in the colour wheel all across the screen..
* re: hacky part in the widget layout. wouldn't it be possible/cleaner to call ` nk_layout_row_dynamic(ctx, cell_ht, 2);` only if the parameter id inside the module is 0 or the previous param is not `colwheel` and then have each widget grab their bounds as customary via `nk_widget_bounds`, without the need to do push right shenanigans?
* offset is essentially lift but after the fact if i understand that correctly? needs documentation in `grade/readme.md` in the parameters section.
* [edit, one more:] the widgets don't have tooltips (these are the strings from the `readme.md` if you wire it like the other widgets)

hanatos:
awesome! this looks workable now, i have some small fixes around this that are semi-unrelated but improve the overall workings, will test a bit more and push if i don't find any more issues :)

PR #266, opened 2026-05-06
grade: tune color wheel effect scaling and sensitivity

Randalix:
Reduce effect to 1% of original strength via shader multiplicator, increase wheel drag sensitivity to maintain responsive control. Adjust colwheel ranges to match actual achievable limits from geometric constraint.

### DIFF
--- a/src/gui/render_darkroom.h
+++ b/src/gui/render_darkroom.h
@@ -567,7 +567,7 @@ render_darkroom_widget(int modid, int parid, int is_fav_menu)
         float dx = ctx->input.mouse.pos.x - ctx->input.mouse.prev.x;
         float dy = ctx->input.mouse.pos.y - ctx->input.mouse.prev.y;
         int shift = nk_input_is_key_down(&ctx->input, NK_KEY_SHIFT);
-        float sens = shift ? 0.1f : 0.01f;
+        float sens = shift ? 0.15f : 0.005f;
         float sx = dx / (radius * 2.0f) * sens;
         float sy = dy / (radius * 2.0f) * sens;
         // constrained so sum of R+G+B deltas stays 0 (chromaticity only)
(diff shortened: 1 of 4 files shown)


hanatos:
..thinking about it maybe it's fine as it was: there will be some place where to multiply a global range to the zero-centered numbers. be it in the shader or in params.ui, it'll always be hidden from the user. so doesn't make much of a difference i suppose.

the one thing i'll probably do is remove the min/max (keep master min/max), and document somewhere that it'll only go up to uhm why is that 0.31? and make sure that the ranges in the grade/params.ui don't clamp away colour range.

alternatively i'd use one max value which would define the radius of the colour wheel, so the rgb values would go up to this radius instead of 0.31.

Randalix:
What if we restore the ranges to +-0.5 so they don't clip the geometric maximum, and document why that 0.31 limit exists? That covers the main issue (no clamping). Does that sound reasonable, or would you prefer the cleaner approach of removing  min/max entirely?

Randalix:
I ended up going the more precise approach: set the ranges to exactly the  geometric limit:  ±0.317 for lift/offset, ±0.683-1.317 for gamma/gain.
This is the maximum per-component displacement given the R+G+B=0 constraint (one component maximized, others balanced symmetrically).
The params.ui constraint now directly matches the geometry—no clipping, no wasted space.
Added documentation in `readme.md` and `render_darkroom.h` explaining the 0.317 limit.

hanatos:
aehem this makes little sense to me.. did you vibe this together :P ? i settled for a variant now that uses -0.5..0.5 as colour range and the master slider can do the rest. the global ui ranges are now *only* for the master slider, so the extra min/max on the params struct went away. i think as a widget that makes a lot of sense now, one gives you colour and the other some master range, do with it what you have to (scale, multiply, add, ..).

PR #268, opened 2026-05-08
dragkeys saturation: use the sat slider in the colour module

maxharicot:
The drag key for saturation seems to use an old version of filmcurv with a chroma slider. Saturation must have moved to the colour module in the meantime, but the key file wasn't updated.

### DIFF
--- a/bin/dragkeys/saturation
+++ b/bin/dragkeys/saturation
@@ -1,3 +1,3 @@
-# hold key and drag mouse to adjust chroma on film curve
+# hold key and drag mouse to adjust saturation in colour
 key:s
-filmcurv:01:chroma:0:0.002
+colour:01:sat:0:0.002


hanatos:
thanks for the pr! i think the film curve/chroma was intended and points to https://github.com/hanatos/vkdt/blob/master/src/pipe/modules/filmcurv/params.ui#L6
the chroma slider in the `filmcurv` module which is only active if in `lum ratio` mode. arguably pointing a drag key that ships with the rest of the default settings in vkdt would better point to the `colour` module version, i think you have a valid point there.

maxharicot:
I see! I always use the agx mode in filmcurv, so I had not seen this slider. The drag keys are very convenient and having it default to a slider that is indeed available in the default graph is sensible.

hanatos:
@Randalix this makes sense to me.. maybe you can repoint the dragkey in your `~/.config` to change the film curve in a way that is consistent with the rest of your default setup in this case :)

PR #269, opened 2026-05-14
hdrmerge: combine multiple images of different exposures to create radiance map and increase dynamic range

phcreery:
I had a project that required more dynamic range that my camera allows for. I found and used a couple tools but though it would be great to add it as a module to vkdt.

This module is essentially a re-implementation of https://github.com/wjakob/hdrmerge

Feedback wanted. I have only tested it with very few sets of images shown below. I would like to do some more testing or see others try it.

What would pair excellently with this module is the align module. Unfortunately, the align module requires the images to be taken at the same exposure.

This requires scene-referred linear raw inputs. I toyed with implementing HRD algorithm by Debevec to calculate the CRF for non linear inputs but I am not smart enough to get the linear algebra least-squares fit in a gpu-centric flow.
https://www.pauldebevec.com/Research/HDR/debevec-siggraph97.pdf. Maybe it can become a cli util?

Discussion at https://discuss.pixls.us/t/hdrmerge-module-feedback-wanted/57685


src/pipe/modules/hdrmerge/linear.comp
```diff
@@ -0,0 +1,174 @@
...
+layout(set = 1, binding = 1) uniform sampler2D img_in2;         // input2
+layout(set = 1, binding = 2) uniform sampler2D img_in3;         // input3
+layout(set = 1, binding = 3) uniform sampler2D img_in4;         // input4
+layout(set = 1, binding = 4) uniform writeonly image2D img_out; // output
+// layout(set = 1, binding = 5) uniform writeonly image2D img_dspy;
+layout(set = 1, binding = 5, r32ui) uniform uimage2D img_dspy;
+
+float triangular_weight(float z)
+{
+  float eps = 0.0001; // 1/pow(10, 4)
```
hanatos: why would a power of ten mean anything here? shouldn't this be more like
* big enough, we don't care
* FLT_EPSILON and FLT_MIN
?

src/pipe/modules/hdrmerge/linear.comp
```diff
@@ -0,0 +1,174 @@
...
+
+void main()
+{
+  ivec2 ipos = ivec2(gl_GlobalInvocationID);
+  if(any(greaterThanEqual(ipos, imageSize(img_out)))) return;
+
+  int n = clamp(params.num_inputs, 1, 4);
+
+  // create array of exposures and corresponding weights
+  float Z[4] = float[](0.0, 0.0, 0.0, 0.0);
```
hanatos: if you can, please avoid arrays in glsl shaders. if the compiler cannot infer the indices compile time (sometimes it's stubborn to unroll loops), it'll spill this into global memory and make it some 20x slower.

src/pipe/modules/hdrmerge/linear.comp
```diff
@@ -0,0 +1,174 @@
...
+  
+  float X_out = E * t_ref;
+  
+  float rgb;
+  if (params.output_mode == 0)      // exposure
+    rgb = X_out;
+  else if (params.output_mode == 1) // radiance
+    rgb = E;
+
+  // // log display irradiance
```
hanatos: are all this comment blocks leftovers from inline dspy mapping?

src/pipe/modules/hdrmerge/main.c
```diff
@@ -0,0 +1,103 @@
...
+  const dt_image_params_t *img3_param = have_input3 ? dt_module_get_input_img_param(graph, module, dt_token("input3")) : 0;
+  const dt_image_params_t *img4_param = have_input4 ? dt_module_get_input_img_param(graph, module, dt_token("input4")) : 0;
+
+  int num_inputs = 0;
+  for(int i=0;i<4;i++)
+    if(module->connector[i].connected.i >= 0)
+      num_inputs++;
+
+  // check if param filters are the same, if not, we can't merge and need to bypass:
+  if(img1_param->filters != img2_param->filters)
```
hanatos: maybe set a gui error message in this case, like here: https://github.com/hanatos/vkdt/blob/master/src/pipe/modules/i-jpg/main.c#L237

src/pipe/modules/hdrmerge/main.c
```diff
@@ -0,0 +1,103 @@
...
+    return dt_connector_bypass(graph, module, 0, 1);
+  if(have_input3 && img1_param->filters != img3_param->filters)
+    return dt_connector_bypass(graph, module, 0, 1);
+  if(have_input4 && img1_param->filters != img4_param->filters)
+    return dt_connector_bypass(graph, module, 0, 1);
+
+  // ensure images are same size
+  if(module->connector[0].roi.wd != module->connector[1].roi.wd ||
+     module->connector[0].roi.ht != module->connector[1].roi.ht)
+    return dt_connector_bypass(graph, module, 0, 1);
```
hanatos: oh, please don't mess with the graph topology in `commit_params` this will be called inline after everything is setup, for every frame. it assumes the graph will stay valid. changing topology/dropping nodes might result in flaky behaviour. call `dt_connector_bypass` only in `create_nodes`.

src/pipe/modules/hdrmerge/map.comp
```diff
@@ -0,0 +1,53 @@
...
+  if(any(greaterThanEqual(ipos, imageSize(img_out)))) return;
+
+  uint cnt = texelFetch(img_in, ipos, 0).r;
+  vec3 rgb = 1.0/255.0*vec3(
+        cnt & 1023,
+       (cnt >> 10) & 1023,
+       (cnt >> 20) & 1023);
+  // rgb = pow(rgb, vec3(0.3));
+
+  float t[4] = float[](params.t1, params.t2, params.t3, params.t4);
```
hanatos: same comment about arrays here. may or may not be fine :)

src/pipe/modules/reinhard/connectors
```diff
@@ -0,0 +1,2 @@
+input:read:rgba:*
+output:write:rgba:f16
```
hanatos: this is of course the memory efficient choice, keeping everything in `f16` here. i'm thinking since this is such a technical module, maybe offering an option as a parameter to pick `f32` here might be one place where this could be useful. anyways this is something that might be discussed/added later. also thinkable that we'd instead introduce a "force format" node later and then convert this to `*` here instead, just thinking out loud.

src/pipe/modules/reinhard/main.c
```diff
@@ -0,0 +1,13 @@
+#include "modules/api.h"
```
hanatos: this whole file is redundant in a simple module like this. you could remove both `flat.mk` and `main.c` (see for instance the `exposure` module) and the core will create all the necessary wires to call `main.comp`.

src/pipe/modules/reinhard/main.comp
```diff
@@ -0,0 +1,63 @@
...
+  float  max_w; // maximum white point for extended Reinhard
+} params;
+layout(set = 1, binding = 0) uniform sampler2D img_in;
+layout(set = 1, binding = 1) uniform writeonly image2D img_out;
+
+// Helper function to calculate relative luminance (Rec. 709 constants)
+float luminance(vec3 v)
+{
+  // return dot(v, vec3(0.2126, 0.7152, 0.0722)); // standard Rec. 709
+  return dot(v, vec3(0.2627, 0.6780, 0.0593)); // rec. 2020 coefficients
```
hanatos: there's a `luminance_rec2020()` function in `shared.glsl`. please use that to keep the spread of matrix coefficients low :)

src/pipe/modules/reinhard/readme.md
```diff
@@ -0,0 +1,15 @@
+# reinhard: tone mapping for hdr input
+
+## parameters
+
+* `max_w` maximum white value
+
+## connectors
+
+* `input`
```
hanatos: i often find it helpful if these connectors contain the future tooltips instructing me to connect linear or display referred data, bayer or not, maybe an example (scene referred after the `colour` module so it's linear rec2020 in this case).

hanatos:
thanks for the PR! i left a few comments inline in the code. some high level comments here:

* please note down the GPLv3 license in the readme.md, if it's derivative work from hdrmerge. not sure it'll be more than the triangle weight, you can probably judge better.
* now i really want alignment that works with images like this :)
* new modules should be wired in `modules/readme.md` in the table of contents of the docs.

da-phil:
Indeed, this is a nice addition!


> * now i really want alignment that works with images like this :)

Sounds like feature descriptor based approach using good old SIFT features? ;) 

Actually, I was just about to implement image alignment for darktable within the RAW HDR merge functionality, because that's such a common use-case for me and working outside of darktable just for a simple HDR merge is annoying. If we get this pipeline running in vkdt, I might have another reason so eventually re-consider vkdt for some of my editing work ;)

hanatos:
> Sounds like feature descriptor based approach using good old SIFT features? ;)

yeah i suppose that's possible. maybe JP's old normalised cross correlation is a better match for the patch-based `align` module in vkdt? https://scribblethink.org/Work/nvisionInterface/vi95_lewis.pdf

sift would probably mean deriving a simple affine transformation instead of non-rigid dense motion. there is the `mv2rot` module in vkdt that attempts to do that from dense motion vectors, but it doesn't work all that well (yet anyways). not sure what a good way to implement sift on gpu would be.

da-phil:
> sift would probably mean deriving a simple affine transformation instead of non-rigid dense motion. there is the `mv2rot` module in vkdt that attempts to do that from dense motion vectors, but it doesn't work all that well (yet anyways). not sure what a good way to implement sift on gpu would be.

This is already where the problems start, finding the right transformation class and not just always going for a full fledged 8 DOF homography for alignment problems which might be  simply solved with a 3 DOF euclidean rigid body transformation. In my experience a default 8 DOF homography works only well if you can prevent overfitting in your solver, where suddenly completely unrelated parameters such as shear or scale are adapted for a case which would have also worked with a 3 DOF euclidean transformation.
We need to consider the following use-cases:
* HDR stacking of exposure stack taken on a tripod or free hand (this can already require a 8 DOF homography due to perspective corrections, depending on the magnitude of the movements between shots and the scene, i.e. having very close and distant elements
* Focus stacking, this might consider weird effects such as focus breathing, which might also need to be modelled by a scale of a more general 6 DOF affine transformation
* Astro image stacking of images which were not taken on an equatorial mount, but a simple tripod (3 DOF or 6 DOF transformation might be enough)

At least those would be the use-cases which I'd like to use vkdt for.

> not sure what a good way to implement sift on gpu would be.

Given how old this algo is, I assume there should be plenty of blue prints available ;)
E.g this one: https://github.com/Celebrandil/CudaSift

hanatos:
that cudasift sounds reasonably simple to convert cu->glsl indeed. the current alignment module actually only works for certain stacks for focus stacking. if you want to attempt this let me know how i can help/if you have any questions.

da-phil:
> that cudasift sounds reasonably simple to convert cu->glsl indeed. the current alignment module actually only works for certain stacks for focus stacking. if you want to attempt this let me know how i can help/if you have any questions.

Not for the time being, unfortunately. Too much other stuff to do unfortunately... I kinda promised to bring HDR image alignment to darktable first 😅

I'm also not sure if a pure sparse feature based approach will lead to good enough alignment, or if we need some dense method like optical flow or ECC for refinement to achieve sub-pixel accuracy and therefore optimal sharpness after the HDR or focus stack step.

src/pipe/modules/hdrmerge/linear.comp
phcreery: no it doesn't, I toyed with this number until I found something decent but epsilon works just as well.

src/pipe/modules/hdrmerge/linear.comp
phcreery: oh, I did not know about that. thanks for tip!

src/pipe/modules/hdrmerge/linear.comp
phcreery: yea, I'll clean them up

src/pipe/modules/hdrmerge/main.c
phcreery: oh, woops! I did not know but that makes sense. Thanks!

src/pipe/modules/reinhard/connectors
phcreery: I changed it to `*`, I am not sure if enforcing a size matters, I could be wrong

src/pipe/modules/reinhard/main.c
phcreery: oh neat! fixed

src/pipe/modules/reinhard/readme.md
phcreery: woops, I think I added some useful descriptions now

src/pipe/modules/hdrmerge/main.c
phcreery: good idea! added

src/pipe/modules/hdrmerge/map.comp
phcreery: okay, do you know if this branching function I did should perform better?

src/pipe/modules/reinhard/main.comp
phcreery: oh, I did not see that guy, I'll use him instead.

src/pipe/modules/hdrmerge/main.c
```diff
@@ -0,0 +1,158 @@
...
+  
+  const dt_image_params_t *img1_param = dt_module_get_input_img_param(graph, module, dt_token("input"));
+  const dt_image_params_t *img2_param = have_input2 ? dt_module_get_input_img_param(graph, module, dt_token("input2")) : 0;
+  const dt_image_params_t *img3_param = have_input3 ? dt_module_get_input_img_param(graph, module, dt_token("input3")) : 0;
+  const dt_image_params_t *img4_param = have_input4 ? dt_module_get_input_img_param(graph, module, dt_token("input4")) : 0;
+  const dt_image_params_t *img5_param = have_input5 ? dt_module_get_input_img_param(graph, module, dt_token("input5")) : 0;
+  const dt_image_params_t *img6_param = have_input6 ? dt_module_get_input_img_param(graph, module, dt_token("input6")) : 0;
+
+  int num_inputs = 0;
+  for(int i=0;i<4;i++)
```
hanatos: `<6` ?

src/pipe/modules/hdrmerge/map.comp
hanatos: yes branching is better here. i mean you'll compute the condition, but in this code as far as i can tell all subgroup threads would take the same branch, so perf impact is limited.

src/pipe/modules/hdrmerge/readme.md
```diff
@@ -0,0 +1,32 @@
...
+* `pltscale` axis scale of dspy plot
+* `output` either exposure or radiance
+* `ev_push` for exposure output mode
+
+## connectors
+
+* `input` the primary input. note, the camera properties will be passed through this input
+* `input2` the secondary input to merge with
+* `input3` (optional) the third input to merge with
+* `input4` (optional) the fourth input to merge with
```
hanatos, line 27: +5  and 6

src/pipe/modules/hdrmerge/main.c
phcreery: Good catch!

src/pipe/modules/hdrmerge/readme.md
phcreery: oh yea, I will add those

src/pipe/modules/readme.md
```diff
@@ -115,6 +115,7 @@ also see the [list of presets shipped with vkdt](../../../doc/howto/presets/pres
 * [mv2rot: estimate rotation + translation from motion vectors](./mv2rot/readme.md)
 * [resize: add ability to resize buffers](./resize/readme.md)
 * [resnet: gmic convolutional neural network](./resnet/readme.md)
+* [hdrmerge: merge images to form hdr radiance map](./resnet/hdrmerge.md)
```
hanatos: `./hdrmerge/readme.md` :)

phcreery: that's embarassing haha

PR #270, opened 2026-05-28
qvk: decouple UI rendering from compute via dedicated async compute queue

niklasiivari:
- Detect and query support for a separate, dedicated Vulkan compute queue family.
- Create the Vulkan logical device with two queues if they belong to different families.
- Allocate shared resources (images/buffers) in VK_SHARING_MODE_CONCURRENT between the compute and graphics queue families when they are connected to graphics presentation nodes (like the display module) or used for thumbnail generation.
- Fall back to previous behaviour when separate queues are not supported by the hardware.

This ensures that the UI (sliders benefit the most) remains responsive during image computation instead of freezing, improving user experience when working with heavy graphs or lower-end devices. 

Edit: Seems GitHub had auto-enabled Copilot code review in my settings, sorry about that.


hanatos:
heya, thanks for the pull request! (yes, hard to disable the ai stuff on this increasingly strange website..)

did you test this on an AMD card? or intel?

theoretically it'll be sufficient to mark the display node input buffers as concurrent, not sure that finer granularity would buy us anything.

one major caveat: some modules (`draw` and `overlay` at this point) use the graphics/raster units, i.e. graph processing will require `vkCmdDraw*` calls and thus a graphics queue, too. this is all nodes with `node->type == s_node_graphics`. did you test with a `draw` node in the loop? maybe it'd be worth having the concurrent execution in case there are no graphics nodes on the graph. the alternative would be converting these to compute shaders, but i don't really want to go there.

src/db/thumbnails.c
```diff
@@ -582,9 +586,9 @@ dt_thumbnails_load_one(
         VK_IMAGE_ASPECT_COLOR_BIT
       | VK_IMAGE_USAGE_TRANSFER_DST_BIT
       | VK_IMAGE_USAGE_SAMPLED_BIT,
-    .sharingMode           = VK_SHARING_MODE_EXCLUSIVE,
-    .queueFamilyIndexCount = 0,
-    .pQueueFamilyIndices   = 0,
+    .sharingMode           = is_concurrent ? VK_SHARING_MODE_CONCURRENT : VK_SHARING_MODE_EXCLUSIVE,
+    .queueFamilyIndexCount = is_concurrent ? 2 : 0,
+    .pQueueFamilyIndices   = is_concurrent ? queue_indices : 0,
```
hanatos: thumbnails are created in two passes. a background thread processes images -> `~/.cache/vkdt/*bc1` (no sharing required, that's the `_cache_one` variant) and then the ui thread will load these bc1 back in, into exclusive gui-queue images (`_load_one` variant). maaaybe loading the bc1 into memory will profit from an async queue, but i believe this would be so fast that it shouldn't trip up the gui rendering much.

src/pipe/graph-run-nodes-allocate.h
```diff
@@ -660,6 +660,25 @@ create_image(
...
+        dt_connector_t *cin = graph->node[n2].connector + i2;
+        if(dt_connector_input(cin) &&
+           cin->connected.i == node - graph->node &&
+           cin->connected.c == c - node->connector)
+        {
+          if(graph->node[n2].name == dt_token("display")) is_concurrent = 1;
+        }
+      }
+    }
+  }
```
hanatos: maybe that's actually `graph->gui_attached`? also might check whether that logic could be merged into `init_display_flags` in `graph-run-modules.h`. as mentioned in the other comment, as soon as there's a `node->type == s_node_graphics` node on the graph this should cause issues.

niklasiivari:
Thank you for the comments! I added another commit addressing the issues. I tested on Intel (Arc 140v iGPU) + linux only, since that is all I have available right now. 

I had not tested the draw/overlay modules with this, and they did end up in an error on the previous version. Now that is also addressed, but of course this fallback behaviour completely removes any benefit that these changes add when those modules are not present. IMO that is still okay, as the UX improvement at least for my use case is quite substantial.

hanatos:
nice! thanks for the fast update. did you run this through the validation layers? i.e. `make debug` and then `vkdt -d qvk`? on nvidia, i'm getting:
```
[qvk] validation layer: vkCmdPipelineBarrier(): srcStageMask (VK_PIPELINE_STAGE_2_FRAGMENT_SHADER_BIT) is not compatible with the queue family properties (VK_QUEUE_COMPUTE_BIT|VK_QUEUE_TRANSFER_BIT|VK_QUEUE_SPARSE_BINDING_BIT) of this command buffer.
The Vulkan spec states: Any pipeline stage included in srcStageMask must be supported by the capabilities of the queue family specified by the queueFamilyIndex member of the VkCommandPoolCreateInfo structure that was used to create the VkCommandPool that commandBuffer was allocated from, as specified in the table of supported pipeline stages (https://docs.vulkan.org/spec/latest/chapters/synchronization.html#VUID-vkCmdPipelineBarrier-srcStageMask-06461)
```
from here:
```
(gdb) bt
#0  vk_debug_callback (severity=VK_DEBUG_UTILS_MESSAGE_SEVERITY_ERROR_BIT_EXT, type=2, callback_data=0x7fffffff3320, user_data=0x0)
    at qvk/qvk.c:101
#1  0x00007fffc694d063 in ??? () at /usr/lib/libVkLayer_khronos_validation.so
#2  0x00007fffc694f6ba in ??? () at /usr/lib/libVkLayer_khronos_validation.so
#3  0x00007fffc5a57fef in ??? () at /usr/lib/libVkLayer_khronos_validation.so
#4  0x00007fffc5d6ac36 in ??? () at /usr/lib/libVkLayer_khronos_validation.so
#5  0x00007fffc5d77601 in ??? () at /usr/lib/libVkLayer_khronos_validation.so
#6  0x00007fffc5e7a19d in ??? () at /usr/lib/libVkLayer_khronos_validation.so
#7  0x000055555568b138 in record_command_buffer (graph=0x5555559413d8 <vkdt+49328>, node=0x7fff78944010, runflag=1)
    at pipe/graph-run-nodes-record-cmd.h:317
#8  0x000055555567c45c in dt_graph_run_nodes_record_cmd
    (graph=0x5555559413d8 <vkdt+49328>, run=4294967295, nodeid=0x7fffffff6c70, cnt=42, module_flags=s_module_request_none)
    at pipe/graph-run-nodes-record-cmd.h:546
#9  0x0000555555674f01 in dt_graph_run (graph=0x5555559413d8 <vkdt+49328>, run=4294967295) at pipe/graph.c:470
#10 0x000055555560d62f in darkroom_enter () at gui/render_darkroom.c:1382
#11 0x00005555556644d3 in dt_view_switch (view=s_view_darkroom) at gui/view.c:40
#12 0x00005555555d2f27 in render_lighttable_center () at gui/render_lighttable.c:430
#13 0x00005555555db573 in render_lighttable () at gui/render_lighttable.c:1493
#14 0x00005555555c1c58 in dt_gui_render_frame_nk () at gui/render.c:254
#15 0x000055555557d22f in dt_gui_render () at gui/gui.c:722
#16 0x0000555555662a82 in main (argc=3, argv=0x7fffffffe0d8) at gui/main.c:337
```

niklasiivari:
Hmm, seems we need some adjustments for nvidia then (and maybe AMD too). For me the output looks like this:

```

> ❯ bin/vkdt -d qvk
[gui] vkdt 1.0.99-354-gfe508ba7 (c) 2020--2026 johannes hanika
[gui] glfwGetVersionString() : 3.4.0 Wayland X11 GLX Null EGL OSMesa monotonic shared
[gui] monitor [0] eDP-1 at 0 0
[gui] vk extension required by GLFW:
[gui]   VK_KHR_surface
[gui]   VK_KHR_wayland_surface
[qvk] validation layer: Copying old device 0 into new device 0
[qvk] validation layer: Copying old device 1 into new device 1
[qvk] validation layer: Copying old device 0 into new device 0
[qvk] validation layer: Copying old device 1 into new device 1
[qvk] validation layer: Copying old device 0 into new device 0
[qvk] validation layer: Copying old device 1 into new device 1
[qvk] validation layer: Copying old device 0 into new device 0
[qvk] validation layer: Copying old device 1 into new device 1
[qvk] dev 0: vendorid 0x8086
[qvk] dev 0: Intel(R) Graphics (LNL)
[qvk] max number of allocations -1
[qvk] max image allocation size 16384 x 16384
[qvk] max uniform buffer range 1073741824
[qvk] dev 1: vendorid 0x10005
[qvk] dev 1: llvmpipe (LLVM 22.1.5, 256 bits)
[qvk] max number of allocations -1
[qvk] max image allocation size 16384 x 16384
[qvk] max uniform buffer range 65536
[qvk] found support for cooperative matrices 8x16x16
[qvk] found support for cooperative matrices 8x16x16
[qvk] found support for cooperative matrices 8x16x16
[qvk] found support for cooperative matrices 8x16x16
[qvk] found support for cooperative matrices 8x16x32
[qvk] found support for cooperative matrices 8x16x32
[qvk] disabling cooperative matrices because 16x16x16 is not supported
[qvk] num queue families: 3
[qvk] picked device 0 with ray tracing and with float atomics and without coopmat support
[qvk] validation layer: vkCreateDevice extension VK_EXT_debug_marker not available for devices associated with ICD /usr/lib64/libvulkan_intel.so
[qvk] queue 0 is idx 0 family 0
[qvk] queue 1 is idx 1 family 1 (async compute)
[qvk] using A2R10G10B10_UNORM_PACK32 and colour space sRGB nonlinear
[qvk] hdr disallowed in config file, set `intgui/allowhdr:1` to enable
[qvk] vulkan present mode: FIFO (vsync)
[gui] no gamepad found
[gui] no display profile file display.eDP-1, using sRGB!
[gui] [menu] loaded 2 entries for 'lighttable'
[gui] [menu] loaded 70 entries for 'darkroom'
[gui] [menu] loaded 2 entries for 'nodes'
[gui] [menu] loaded 0 entries for 'files'
[qvk] using A2R10G10B10_UNORM_PACK32 and colour space sRGB nonlinear
[qvk] hdr disallowed in config file, set `intgui/allowhdr:1` to enable
[qvk] vulkan present mode: FIFO (vsync)
[qvk] using A2R10G10B10_UNORM_PACK32 and colour space sRGB nonlinear
[qvk] hdr disallowed in config file, set `intgui/allowhdr:1` to enable
[qvk] vulkan present mode: FIFO (vsync)
[qvk] using A2R10G10B10_UNORM_PACK32 and colour space sRGB nonlinear
[qvk] hdr disallowed in config file, set `intgui/allowhdr:1` to enable
[qvk] vulkan present mode: FIFO (vsync)
```

Will try to find out what needs to be done.

hanatos:
did you enter darkroom/edit mode? maybe different version of the validation layers? apparently the pipeline barriers need to include only the stage flags that are compatible with the queue. i.e. the IMG_LAYOUT() macro goes through the IMAGE_BARRIER and then calls into `vkCmdPipelineBarrier` that includes fragment shader bits as `VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT`, just in case. the compute-only pipeline can't do that.

continuing on from there, i have
```[qvk] validation layer: vkCmdPipelineBarrier(): pImageMemoryBarriers[0] barrier using VkImage 0xad00000000ad0[filmsim_part2h_output@0,0] created with sharingMode VK_SHARING_MODE_CONCURRENT, has srcQueueFamilyIndex 0. Source queue family must be VK_QUEUE_FAMILY_IGNORED or VK_QUEUE_FAMILY_EXTERNAL.
The Vulkan spec states: If the synchronization2 feature is not enabled, and image was created with a sharing mode of VK_SHARING_MODE_CONCURRENT, srcQueueFamilyIndex must be VK_QUEUE_FAMILY_IGNORED or VK_QUEUE_FAMILY_EXTERNAL (https://docs.vulkan.org/spec/latest/chapters/synchronization.html#VUID-VkImageMemoryBarrier-None-09053)
[qvk] validation layer: vkCmdBlitImage(): called in VkCommandBuffer 0x55555bf121e0 which was allocated from the VkCommandPool 0xa0f0000000a0f which was created with queueFamilyIndex 2 which contains the capability flags VK_QUEUE_COMPUTE_BIT|VK_QUEUE_TRANSFER_BIT|VK_QUEUE_SPARSE_BINDING_BIT (but requires VK_QUEUE_GRAPHICS_BIT).
The Vulkan spec states: The VkCommandPool that commandBuffer was allocated from must support VK_QUEUE_GRAPHICS_BIT operations (https://docs.vulkan.org/spec/latest/chapters/copies.html#VUID-vkCmdBlitImage-commandBuffer-cmdpool)
[qvk] validation layer: vkCmdBlitImage(): called in VkCommandBuffer 0x55555bf121e0 which was allocated from the VkCommandPool 0xa0f0000000a0f which was created with queueFamilyIndex 2 which contains the capability flags VK_QUEUE_COMPUTE_BIT|VK_QUEUE_TRANSFER_BIT|VK_QUEUE_SPARSE_BINDING_BIT (but requires VK_QUEUE_GRAPHICS_BIT).
The Vulkan spec states: The VkCommandPool that commandBuffer was allocated from must support VK_QUEUE_GRAPHICS_BIT operations (https://docs.vulkan.org/spec/latest/chapters/copies.html#VUID-vkCmdBlitImage-commandBuffer-cmdpool)
...
```

and then the window freezes.

niklasiivari:
Another fix commit added. Hopefully this resolves the issues also on other hardware. I guess my gpu driver is somewhat more forgiving with those, since I did not face any freezes in darkroom with the previous version either. The vulkan validation errors did come up in darkroom previously though. Now that is also completely clean for me.

src/pipe/modules/mipmap/main.comp
```diff
@@ -0,0 +1,12 @@
...
+layout(set = 0, binding = 0) uniform sampler2D img_in;
+layout(set = 0, binding = 1) uniform writeonly image2D img_out;
+void main() {
+    ivec2 pos = ivec2(gl_GlobalInvocationID.xy);
+    ivec2 size = imageSize(img_out);
+    if(pos.x >= size.x || pos.y >= size.y) return;
+    vec2 uv = (vec2(pos) + 0.5) / vec2(size);
+    vec4 color = texture(img_in, uv);
+    imageStore(img_out, pos, color);
+}
```
hanatos, line 12: this might be a prime use case for `svgf2/down.comp` to create three levels of mip maps at the same time in very fast. not sure the subgroup stuff works on intel though, might be worth checking.

src/pipe/graph-run-nodes-allocate.h
```diff
@@ -331,7 +331,7 @@ bind_buffers_to_memory(
   { // storage buffer
     if(img->image_view) vkDestroyImageView(qvk.device, img->image_view, VK_NULL_HANDLE);
     img->image_view = 0;
-    QVKR(vkBindBufferMemory(qvk.device, img->buffer, graph->vkmem, img->offset));
+    QVKR(vkBindBufferMemory(qvk.device, img->buffer, graph->vkmem[img->mem_type], img->offset));
```
hanatos, line 344: ouch now i have some merge conflicts with this and my local master (messed with memory allocations for another reason). give me a bit to think about how to resolve this..

src/pipe/graph-run-nodes-allocate.h
```diff
@@ -1112,53 +1135,32 @@ dt_graph_run_nodes_allocate(
...
-      vkFreeMemory(qvk.device, graph->vkmem_staging, 0);
-      graph->vkmem_staging = 0;
+      *run |= s_graph_run_upload_source; // new mem means new source
+      if(graph->vkmem[i])
+      {
+        vkFreeMemory(qvk.device, graph->vkmem[i], 0);
+        graph->vkmem[i] = 0;
+      }
+      graph->vkmem_size[i] = 0;
+      QVKR(dt_check_device_allocation(graph->heap[i].vmsize, 0));
```
hanatos: this allocation check needs to pass 0 for the device local memory and 1 for staging memory (host visible)

src/pipe/graph-run-nodes-record-cmd.h
```diff
@@ -22,74 +22,80 @@ generate_mipmaps(
   int wd = img->wd > 0 ? img->wd : rwd;
   int ht = img->ht > 0 ? img->ht : rht;
 
-  // put all mip levels into transfer_dst_optimal, since we want to blit onto them:
+  // put all mip levels into general layout since we will read/write them with compute:
```
hanatos: in fact yes that's one of the patches i have locally, use the vk extension that treats all image layouts the same, so we can simply pass GENERAL in all cases.

hanatos:
great update! i have some local changes that may interfere (see code comments above). one is that i have multiple vkmem heaps (one for regular, one for protected memory that is never aliased, one for overflow because ssbo addressing in spirv can't do >32bits without extra tricks, the staging, and now the extra heaps for your use case).

why do we need a separate vkmem allocation for a different queue family?

anyhow thanks for cleaning out the validation layer complaints. i can confirm that your pr (without my local changes) runs clean on nvidia now too (including draw modules).

a fast compute shader for mipmaps can be found in `svgf2/down.comp`. it follows this amd idea to use subgroup ops to output three (2x2 4x4 8x8) levels of mipmaps directly. might be worth wiring instead of yours. unfortunately i don't remember whether it'll work with non-32-wide subgroups (as you'd have on intel).

src/qvk/qvk.c
```diff
@@ -414,13 +427,14 @@ qvk_init(const char *preferred_device_name, int preferred_device_id, int window,
     VK_KHR_GET_MEMORY_REQUIREMENTS_2_EXTENSION_NAME,
     VK_KHR_DEDICATED_ALLOCATION_EXTENSION_NAME,
     VK_KHR_BUFFER_DEVICE_ADDRESS_EXTENSION_NAME,
+    VK_KHR_PUSH_DESCRIPTOR_EXTENSION_NAME,
```
hanatos: probably in the wrong place for non-raytracing devices (needs to bubble up such that the len = 2 case below still includes it)

niklasiivari:
Added two more commits. Hopefully merging the memory stuff won't cause too many difficulties! 

Regarding the vkmem heaps, we don't really need a separate vkmem allocation for the queue family itself, I just switched to allocating the heaps based on memory type indices. Now we can just dynamically calculate the required index based on the memory properties the resources ask for.

hanatos:
okay merged this now. i totally expect some weird bugs to surface.. these are quite a few deep core graph changes at the same time.

anyhow, if you look at the diff you'll notice that i didn't merge the memory type heaps. i can see value in having an array of heaps, maybe that's something to refactor again in the future. these heaps facilitate memory aliasing/reusing graph memory that isn't required by the nodes any more. doing that brings a graph with some medium sized image and llap and oidn on the graph down from like 45GB to 4GB memory usage, so it's kinda essential. then i need extra heaps for memory that can't be aliased (used in animations) and host visible staging. so the heaps don't map 1:1 to gpu heaps.

i'll push the combined mess now and hope for the best :) will do some testing on lower end intel too.

PR #271, opened 2026-05-28
gui: support Ctrl+scroll slider adjustments and optimize dynamic LOD transitions

niklasiivari:
- Implement Ctrl+scroll support for floats, integers, and multi-component slider widgets.
- Add translation mappings for GLFW left/right control keys to NK_KEY_CTRL in the Nuklear keyboard callback, allowing Nuklear to detect Ctrl modifiers.
- Introduce a CONSUME_SCROLL macro in render_darkroom.h to adjust the parameter value, zero out the scroll delta to prevent Nuklear background window scrolling, and activate dynamic LOD.
- Centralize dynamic level-of-detail (LOD) handling inside render_darkroom() using interact_begin/interact_end events and a scroll_busy frame countdown.
- Auto-release active slider widgets at frame boundaries when the mouse is released to prevent unresponsiveness or focus leak.
- Switch parameter-change runflags from direct assignment (=) to bitwise OR (|=) to accumulate rather than discard multiple updates in the same frame.

### DIFF
--- a/src/gui/gui.h
+++ b/src/gui/gui.h
@@ -83,6 +83,7 @@ typedef struct dt_gui_wstate_t
   int      active_widget_parid; // parameter id
   int      active_widget_parnm; // number of parameter (if multi count)
   int      active_widget_parsz; // size of parameter to copy
+  int      active_widget_is_slider;
   struct nk_rect active_dspy_bound;
   int      selected;
   float    aspect;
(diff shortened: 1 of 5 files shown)


hanatos:
nice! works mostly as intended. one small thing: if you run with dynamic LOD=4 or something and move a slider in the node editor, now it stays at 100% resolution. current master would also apply lod here.

i suppose it's as simple as replicating some of the preamble in `render_darkroom()` in `render_nodes()` too.

niklasiivari:
Good catch! I have probably never used the sliders in the node editor view, so this totally slipped past my testing. Fixed now.

PR #272, opened 2026-05-29
filmsim: performance optimizations, better grain, preflash

niklasiivari:
One more PR from me. This aims to improve the filmsim module performance by optimizing the math while achieving output that is visually indistinguishable. 

Note: Based on my testing, the vectorized newton solver does cause a small divergence in how large coupler values affect the image. This may even be a beneficial side effect as it seems to reduce a warming effect that was present previously.

In addition, swapped the grain model for a faster and at least in my opinion visually nicer implementation that is on some level based on the Spektrafilm implementation, just built in a way that it can be computed fast on a gpu.  

While the code works well for me and significantly improves performance, I am initially opening this as a draft as this will probably need some adjustments for code clarity, and maybe we can even squeeze a few more milliseconds off somewhere. 

I am aware this will likely make future additions and improvements somewhat more difficult, but perhaps that is a reasonable tradeoff if this gives good results also on other hardware. Interested to hear how it performs.

Comments and improvement suggestions very much welcome!

# Benchmark results:

Setting
--
- 100 iterations (10 first treated as warmup)
- input: 40Mp Fujifilm x-trans compressed RAW
- config based on default filmsim preset, but grain and halation enabled
- Intel Arc 140v integrated laptop gpu with 32gb memory-on-package

Results
--
Here the `total time` is the most accurate metric as the others seem to depend a lot on the GPU scheduling as we can see from the fact that sum filmsim fluctuates from min 0.019 ms and max 287 ms on master. 

The Maximum times of the different stages should give a pretty good picture of what their actual contribution is though.

> Only a subset of the metrics is displayed for readability

Metric | Stat | master (ms) | filmsim-perf (ms) | Diff (ms) | % Change
-- | -- | -- | -- | -- | --
filmsim part0 | avg | 2.485 | 0.542 | -1.944 | -78.2%
filmsim part0 | min | 0.002 | 0.002 | +0.000 | +0.0%
filmsim part0 | max | 56.000 | 12.163 | -43.837 | -78.3%
filmsim part1h | avg | 1.285 | 0.003 | -1.282 | -99.8%
filmsim part1h | min | 0.002 | 0.002 | +0.000 | +0.0%
filmsim part1h | max | 23.228 | 0.003 | -23.225 | -100.0%
filmsim part2h | avg | 73.819 | 0.773 | -73.046 | -99.0%
filmsim part2h | min | 0.003 | 0.002 | -0.001 | -33.3%
filmsim part2h | max | 246.092 | 34.678 | -211.414 | -85.9%
shared blurh | avg | 1.410 | 0.076 | -1.334 | -94.6%
shared blurh | min | 0.002 | 0.002 | +0.000 | +0.0%
shared blurh | max | 37.063 | 14.059 | -23.004 | -62.1%
shared blurv | avg | 1.266 | 0.293 | -0.973 | -76.9%
shared blurv | min | 0.002 | 0.002 | +0.000 | +0.0%
shared blurv | max | 36.546 | 13.908 | -22.638 | -61.9%
sum filmsim | avg | 82.157 | 1.946 | -80.210 | -97.6%
sum filmsim | min | 0.019 | 0.017 | -0.002 | -10.5%
sum filmsim | max | 287.929 | 34.693 | -253.236 | -88.0%
total time | avg | 505.623 | 175.527 | -330.097 | -65.3%
total time | min | 496.386 | 172.999 | -323.387 | -65.1%
total time | max | 516.321 | 178.477 | -337.844 | -65.4%

# Grain example

These images were created from the same input image, with the same config.

Old:
Screenshot from 2026-05-30 17-12-48

New:
Screenshot from 2026-05-30 17-13-44


hanatos:
wow really nice perf improvements :)

do you have the "original" spektrafilm version of the grainy image too? it does look substantially different, especially less colourful and with suppresed low frequency content. more like single-octave simplex noise, something that looks CG to my eye ;)

niklasiivari:
Hi, I can see what you mean with the grain, I'll take another look and maybe come up with something better!

niklasiivari:
Pushed an update with a better noise model now a lot more closely aligned with the spektrafilm implementation. It is a bit slower than the previous version, but also a lot better looking. Uploaded new comparison images to the PR description. The differences between the original vkdt implementation are not huge, but in my opinion still visible, and this seems to be faster than the original approach. 

A quick benchmark of the new version gives:
```
 total time      : Avg  218.969 ms  (Min:  217.307 ms, Max:  220.904 ms) [collected 100 frames]
```

niklasiivari:
Regarding the preflash commit, hopefully this isn't pushing too much into scope creep. I just saw that there were comments in the shader code concerning this, and since I was already working on that code, thought I would implement that here too to avoid merge conflicts from another pr.

src/pipe/modules/filmsim/part2h.comp
```diff
@@ -10,19 +10,27 @@ layout(push_constant, std140) uniform push_t
...
+  // We must initialize before the boundary check so all threads execute the barrier.
+  if(params.process == 0)
+    init_enlarger_shared(params.film, s_paper_offset + params.paper);
+  init_scan_shared();
+
   ivec2 ipos = ivec2(gl_GlobalInvocationID);
   if(any(greaterThanEqual(ipos, imageSize(img_out)))) return;
   // load halation + raw exposed film
-  vec3 raw = texture(img_in,  (ipos+0.5)/vec2(imageSize(img_out))).rgb;
-  vec3 hal = texture(img_hal, (ipos+0.5)/vec2(imageSize(img_out))).rgb;
```
hanatos, line 17: i hope this is not one of the places where texture() would do the interpolation for the upsampling during enlargement? does the 2x mode still work?

src/pipe/modules/filmsim/filmsim.glsl
```diff
@@ -272,105 +546,67 @@ develop_film(vec3 log_raw, int film, ivec2 ipos, float scale)
...
-    // lamp filters are transmittances 0..100%
-    vec3 enlarger = 100.0*mix(
-      vec3(1.0),
-      vec3(sigmoid_eval(coeff_c, lambda), sigmoid_eval(coeff_m, lambda), sigmoid_eval(coeff_y, lambda)),
-      neutral);
-#endif
-    float print_illuminant = enlarger.x*enlarger.y*enlarger.z * illuminant;
-    float light = pow(10.0, -density_spectral) * print_illuminant;
-    raw += sensitivity * light * pow(2.0, params.ev_paper);
-    // TODO and the same yet again for the preflash
```
hanatos, line 370: whoaah preflash! :)

hanatos:
new noise looking great!

src/pipe/modules/filmsim/part2h.comp
hanatos: answering myself: seems to work.

niklasiivari: Yes, it still works! The upsample interpolation seems to be done in part1, as there i needed to use texture (made the mistake of switching it there too initially, but noticed that 2x/4x broke so reverted later). Here texelFetch works just fine!

e: didn't see your response as I was just typing :)

hanatos:
just gave it a spin on an nvidia card (rtx4080s), 20MP image:
```
master:
[perf] colour   main    :	   0.362 ms
[perf] filmsim  part0   :	   3.481 ms
[perf] shared   blurh   :	   1.340 ms
[perf] shared   blurv   :	   0.629 ms
[perf] filmsim  part1h  :	   1.605 ms
[perf] shared   blurh   :	   0.980 ms
[perf] shared   blurv   :	   1.004 ms
[perf] filmsim  part2h  :	  12.584 ms
[perf] sum filmsim:	  21.622 ms

pr:
[perf] colour   main    :	   0.361 ms
[perf] filmsim  part0   :	   0.690 ms
[perf] shared   blurh   :	   0.617 ms
[perf] shared   blurv   :	   0.781 ms
[perf] filmsim  part1h  :	   0.530 ms
[perf] shared   blurh   :	   0.969 ms
[perf] shared   blurv   :	   0.962 ms
[perf] filmsim  part2h  :	   3.883 ms
[perf] sum filmsim:	   8.433 ms
```

really really cool :) instant merge, will read the code in detail later when i merge in the spectral skin prior upsampling..

PR #273, opened 2026-05-30
Update flake to bring in cargo fetch fixes

jayrahdevore:
crates.io started blocking generic requests, which broke the nix fetcher.  Update the flake.lock to bring in fixes that add a proper user agent string to requests.

See https://github.com/rust-lang/crates.io/issues/13482

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1774386573,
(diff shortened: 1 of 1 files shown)


hanatos:
thanks for this! i can't say i understand the issue.. seems to work here before and after merging. i suppose so it doesn't hurt. pushed!

jayrahdevore:
Thanks!  It probably wouldn't cause any errors if the crates were already downloaded in a nix store (no fetch needed), but it would cause an issue for new users and those updating from an older version of rawler

PR #275, opened 2026-05-30
Fix crashes on devices that do not have VK_EXT_shader_64bit_indexing or VK_KHR_unified_image_layouts support

niklasiivari:
Latest master was crashing on my machine due to requiring the VK_EXT_shader_64bit_indexing support unconditionally. Made it optional and loaded only if its supported.

Also made the VK_KHR_unified_image_layouts optional, since it isn't supported by a lot of older hardware.

### DIFF
--- a/src/gui/nuklear_glfw_vulkan.h
+++ b/src/gui/nuklear_glfw_vulkan.h
@@ -520,7 +520,7 @@ nk_glfw3_create_pipeline(struct nk_glfw_device *dev)
     };
     VkGraphicsPipelineCreateInfo pipeline_info = {
       .sType = VK_STRUCTURE_TYPE_GRAPHICS_PIPELINE_CREATE_INFO,
-      .pNext = &pipeline2_info,
+      .pNext = qvk.shader64bit_indexing_supported ? (void*)&pipeline2_info : 0,
       .flags = 0,
       .stageCount = 2,
       .pStages = shader_stages,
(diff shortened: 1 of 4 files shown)


hanatos:
hah, nice. thanks for cleaning up this messy part of the vulkan init.

so, about the 64 bit ext. maybe it's actually a bad idea. i don't think it helps me extend the ssbo/storage image addressing range since vkdt has more the use case of very many small image buffers instead of one with base address < 4G that extends outside due to large buffer size. also doubling register pressure for address calculations may be a bad idea. so i guess i'm saying i might disable this extension again in the future.

about image layouts: seems nvidia 20xx supports it, 10xx doesn't. in general more like 10% of current gpus, says gpuinfo.org :(

the problem is that simply disabling the extension and leaving all image layouts at "general" might have a perf impact on older devices. might have been a bit fast pushing this. do you notice a speed impact on your arc 140v (was it that?)?

the thing is that the image layout transitions in a graph with multiple i/o had edge cases and issues, so the old version wasn't 100% correct.

niklasiivari:
No worries!

I can't really comment on the impact of the "general" layouts since my GPU does have support for that unified layouts extension. I guess the 64bit indexing extension wasn't supported because it also has a 4GB limit for the buffers anyway, despite being quite new.

hanatos:
ah the patch with the unified_image_layouts came with a lot of replacement SHADER_READ_ONLY_OPTIMAL->GENERAL which would be sub-optimal and potentially slow. that's the part i mean.

niklasiivari:
Sorry, meant that since the performance impact should be mostly only on older gpu's that do not have support for that extension, I did not really notice anything.

PR #278, opened 2026-06-02
jddcnn: fix intel (and maybe AMD too)

niklasiivari:
Wanted to try out the jddcnn module now that it also supports x-trans. However, quickly noticed that it was not working on my gpu at all. This module seems to still be work-in-progress, so not sure if you want to integrate this kind of changes, but just putting this out here in case it is useful to you or someone else trying to get this to work!

### DIFF
--- a/src/pipe/graph-run-nodes-allocate.h
+++ b/src/pipe/graph-run-nodes-allocate.h
@@ -1101,18 +1101,17 @@ alloc_outputs(dt_graph_t *graph, dt_node_t *node)
 
       // TODO: cache pipelines on module->so ?
       // vk 1.3:
-      // VkPipelineShaderStageRequiredSubgroupSizeCreateInfoEXT sub = {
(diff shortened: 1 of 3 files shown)


hanatos:
nice. interesting, why do you need buffer address refs for intel? maybe in general now because i bind them? the spirv for the neural network doesn't use the buffer addresses, fwiw, it goes through ssbo bindings.

the "generic" codepath here is just some nvidia 30xx card without cooperative matrices, so it seems plausible that fixing subgroup sizes would be required.

the actual way to support intel/amd for real is of course to compile/autotune the onnx once more for these devices, see https://codeberg.org/hanatos/vkdt-nn . give us a few days to release the compiler code..

niklasiivari:
Not completely sure why those are needed, somehow I just couldn't get it to work without. 

Sounds like it's better to not integrate this change right now and instead just wait for the AMD/intel optimized shaders, so I'll close this PR now.

PR #281, opened 2026-06-23
gui: add color assessment mode

niklasiivari:
This pull request adds a darktable-like color assessment mode with customizable margin and border sizes.

image

### DIFF
--- a/src/gui/gui.c
+++ b/src/gui/gui.c
@@ -220,6 +220,9 @@ int dt_gui_init()
   if(snprintf(configfile, sizeof(configfile), "%s/config.rc", dt_pipe.homedir) < 512)
     dt_rc_read(&vkdt.rc, configfile);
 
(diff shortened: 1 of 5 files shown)


hanatos:
seems to work really well, thanks!

WillPower3309:
I think we should make the outer light grey area the full background for vkdt - we could make the sidebar darker to differentiate it. This would make it a bit easier to set exposure outside of this mode as well

hanatos:
some customisation is possible via `style.txt`: changing
```
WINDOW:200:20:20:255
```
makes the panel red
```
DT_BACKGROUND:70:7:7:255
```
renders the center background red, for instance. the `DT_*` are vkdt extensions, the others are plain nuklear widget colours.

you can hide the panel in dr mode by pressing `<tab>`. this will also make the border go away. you can still use the chord menu or the right click/gamepad mode to adjust exposure in this view.

you should be able to put your custom `style.txt` into `~/.config/vkdt/` to override the shipped version (copy the whole file, it will only read one of them).

PR #284, opened 2026-07-15
filmsim: use subgroupadd for autoexp/preflash

niklasiivari:
Atomic add in shared memory was not supported on my device. Fixed with a fairly universal implementation based on a combination of subgroupadd and loops.


hanatos:
you're awesome :) thanks for the quick fix! i wasn't sure these atomics were widely supported.. gpuinfo.org seemed to indicate all vendors' recent GPUs were fine, obviously not enough. way more elegant this way anyways.

PR #286, opened 2026-07-27
widget_image: proportional zoom with configurable gui/zoom_speed

qosch:
Makes scrolling in the darkroom smoother when using a touchpad (fractional scrolling) and adds a setting to modify speed. It now feels much snappier on my laptop.

### DIFF
--- a/src/gui/widget_image.h
+++ b/src/gui/widget_image.h
@@ -321,8 +321,7 @@ dt_image_events(struct nk_context *ctx, dt_image_widget_t *w, int hovered, int m
     if(yoff != 0)
     { // mouse wheel (yoff is the input)
       ctx->input.mouse.scroll_delta.y = 0;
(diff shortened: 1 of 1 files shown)


hanatos:
thanks! will have a look. does it still feel natural with a mouse wheel?

qosch:
This does two things:
1. when scroll increments come in faster than they are processed (yoff >1), they aren't missed, but the magnification is incremented in the resulting larger step
2. for touchpads that support fractional scrolling, increments <1 can happen. These should now be processed in a proportional way (half a scroll increment gives half the magnification step)
1 should improve behavior both with touchpads and scroll wheels, 2 should not have any impact on behavior with scroll wheels (as they only send integer increments).

An LLM claims to have found another issue concerning updating of thumbnails during scrolling which I am unable to judge, but I'll open a ticket to discuss that

PR #287, opened 2026-07-28
colour, clut: add fixed-grid multi-anchor profiles and auto-temperature

niklasiivari:
After getting something of an approximate input device transform for my camera working, I was kind of wishing for a way to actually white balance using the temp slider. Ended up feeling restricted by the dual illuminant interpolation, so implemented multi-anchor clut generation, and added support for those in the colour module, extended slider range to cover more temperatures, and added an auto temperature calculation that estimates the image cct from the raw metadata by finding what clut evaluation temperature would give the most neutral output. Additionally, can use a picked patch as a temp target by choosing temp as picked target and setting temp to 0. 

Everything works with older dual-illuminant luts too.

### DIFF
--- a/src/tools/clut/src/mkclut.c
+++ b/src/tools/clut/src/mkclut.c
@@ -7,6 +7,121 @@
 #include "core/fs.h"
 #include "core/half.h"
 #include "core/inpaint.h"
+#include <math.h>
+
+#define MKCLUT_MAX_ANCHORS 16
+
+// Planckian spectrum, 360--830nm at 1nm.
+static inline int
+synth_illuminant_planckian(
+    double         T,
+    double       (*out)[4])
+{
+  const double h = 6.62606957e-34; // planck's constant [J s]
+  const double c = 299792458.0;    // speed of light [m/s]
+  const double k = 1.3807e-23;     // boltzmann's constant [J/K]
+  int cnt = 0;
+  for(double l = 360.0; l <= 830.0; l += 1.0, cnt++)
+  {
+    const double lambda_m = l * 1e-9;
+    const double lambda5 = lambda_m*lambda_m*lambda_m*lambda_m*lambda_m;
+    const double c1 = 2.0 * h * c * c / lambda5;
+    const double c2 = h * c / (lambda_m * T * k);
+    out[cnt][0] = l;
+    out[cnt][1] = c1 / (exp(c2) - 1.0);
+    out[cnt][2] = out[cnt][3] = 0.0;
+  }
+  return cnt;
+}
+
+// CIE daylight basis functions, 300--830nm at 5nm.
+static const double cie_day_wl[107] = {
+  300,305,310,315,320,325,330,335,340,345,350,355,360,
+  365,370,375,380,385,390,395,400,405,410,415,420,425,
+  430,435,440,445,450,455,460,465,470,475,480,485,490,
+  495,500,505,510,515,520,525,530,535,540,545,550,555,
+  560,565,570,575,580,585,590,595,600,605,610,615,620,
+  625,630,635,640,645,650,655,660,665,670,675,680,685,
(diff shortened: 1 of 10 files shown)


hanatos:
very nice! also the auto-wb looks really solid. still have to review the code in detail, i assume the sink download would be skipped / dead code eliminated on the graph once the coefficients are found or the auto mode is disabled?

i suppose the next step in this line of thought would be store just a single lut from camera rgb -> spectral upsampling coefficients and then do all the white balancing run-time by dividing some scene illuminant in the spectral domain (see the other pr..). i was under the impression that this would be way too slow, hence the lut approach taken here. i still think speed is a valid argument, but once the multi-illuminant-luts grow out of hand a generic solution might be justified for generality.

niklasiivari:
Just did an experiment on this single lut approach and with a quick benchmark it's around 4x slower... Also, if I understand correctly, we might actually be trading some metamerism-"accuracy" if we just encode one camera->spectral transition vs fitting separate ones at different illuminants?

hanatos:
ah, yes i think i forgot the illuminant in this idea. to be able to give a spectrum from camera rgb we'd need to know the illuminant first. so not much can be gained here in terms of generality.

niklasiivari:
I guess we could always do what mkclut does at runtime for the chosen illuminant just from the ssf. But this would probably be quite slow.

src/tools/clut/src/mkclut.c
```diff
@@ -82,122 +195,105 @@ create_chroma_lut(
   const double white_rec2020_L1 = normalise1(rec2020);
   const double ref_L = white_cam_rgb_L1 / white_rec2020_L1;
 
-  // first pass: get rough idea about max deviation from white and the saturation we got there
```
hanatos, line 85: what was your reasoning when removing the first pass here? it's not obvious from the code.. but this is supposed to filter metameric ambiguity by filtering by saturation. the idea is roughly that we can't avoid metamers between observer rgb and camera rgb. that means the mapping will collapse to a non-bijection: there will be observer rgb values that can be reached by two or more different camera rgb values. to stay safe, we pick the ones with the lowest saturation (using a pretty abritrary definition of saturation).

if i understand your code correctly you'd just pick the first one to be in a lut pixel?

attaching an image of an old writeup on the topic, in the hope that it'll make sense to you:
Image
in this top right image with insets you can see how the space "folds onto itself". we'd want to systematically filter these values such that the interpolated version (top right) will end up being smooth.

have you looked at luts generated by your method? can you see artifacts of this kind?

hanatos: looking at my old lut and one created from this branch:

old:
Image
new:
Image

i'll be hard pressed to say which one is better :D the new one supersamples a lot i believe and thus takes much longer than i remember the old one did. it does show some of the speckle noise i'd expect when removing the saturation filtering though.

hanatos:
i can see the abovementioned artifacts even more pronounced in a lut with more anchors for extreme temperatures.

the rest (temperature picking, interpolation, etc..) works really well in my tests!

maybe it's possible to parallelise mkclut over anchor points? maybe using `core/threads.h`, or if really need be with openmp (don't like the extra dependency). but really it doesn't matter much, creating a lut is such a one-time thing if you're not developing the utility :)

src/tools/clut/src/mkclut.c
niklasiivari: I think I agree with old approach being better, pushed a commit restoring that saturation filtering.

hanatos:
better! i think now this looks at least as good as the old version was. somehow i can still see some residual noise, but maybe the disambiguation criterion just doesn't work so well on the camera i'm testing. i'd say that's not a regression but might profit from a second look at some point. again, i didn't spot an issue in actual image rendering so far.

PR #288, opened 2026-07-28
pipe: allocate graph resources from compatible arenas

niklasiivari:
I seem to be hitting some memory allocation errors all the time, so this pr moves memory allocation into a more generalisable pattern. Should probably be tested on other hardware too before possible merge, as I am only able to test on one platform. 

This time the conflict was in ssbo memory (while working on some filmsim stuff), but thought it would be better to fix this more generally instead of adding workarounds.

### DIFF
--- a/src/pipe/alloc.h
+++ b/src/pipe/alloc.h
@@ -15,6 +15,7 @@ typedef struct dt_vkmem_t
   struct dt_vkmem_t *prev;  // for alloced/free lists
   struct dt_vkmem_t *next;
   struct dt_vkalloc_t *heap;// pointing back to owner
+  struct dt_graph_memory_t *memory; // Vulkan arena backing this allocation
 }
 dt_vkmem_t;
 
--- a/src/pipe/graph-run-nodes-allocate.h
+++ b/src/pipe/graph-run-nodes-allocate.h
@@ -327,11 +327,7 @@ bind_buffers_to_memory(
   //     dt_token_str(node->module->inst),
   //     dt_token_str(node->kernel),
   //     img->offset, img->offset + img->size);
-  VkDeviceMemory mem = graph->vkmem;
-  if(img->mem->heap == &(graph->heap_protected))
-    mem = graph->vkmem_protected;
-  else if(img->mem->heap == &(graph->heap_1))
-    mem = graph->vkmem_1;
+  VkDeviceMemory mem = img->mem->memory->vkmem;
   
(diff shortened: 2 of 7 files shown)


hanatos:
have to do some testing here, but looks straight forward on first glance.

fwiw these multi-heaps will keep the base addresses of ssbo below the 32-bit limit. for very large buffers apparently the way to do it is via sparse binding and our own page table. i really hope that won't be necessary because i think that sucks in terms of code complexity.

i have a local branch that works with buffer references instead of ssbo. well in fact it copies buffer references in addition to the ssbo bindings so i don't have to change all the shader kernels at once. buffer references don't have these limitations and might be the clean solution at some point. (and unfortunately that's part of a larger patch which i'm unlikely to merge soon)

niklasiivari:
About this one, if this seems like something that will just complicate your bigger ongoing work, I can also prepare a smaller patch that makes things work for me while keeping things largely similar, should have such sitting on some local branch of mine

The buffer reference approach sounds interesting!

hanatos:
no worries. i think it's reasonably disjoint. exposing buffer addresses to glsl is just one additional last step, the memory allocation itself is the same/can be performed more carelessly. i can try and factor out this thing, shouldn't be all that large just the address part.

hanatos:
hah! it's just too hot here for my brain. i had already merged the buffer address reference part, so we're good :)

mainly the ray tracing uses buffer references because i thought that's where the heavy buffers would be required, exclusively:
https://github.com/hanatos/vkdt/blob/master/src/pipe/modules/rt/raytrace.glsl

and there's one case where i needed extended memory as a hotfix for a rendering project where the last module in the graph was rendering a text overlay but finally exceeded the 32-bit address range:
https://github.com/hanatos/vkdt/blob/master/src/pipe/modules/overlay/main.vert#L18

and that showcases the generic way of accessing memory via buffer address references instead of ssbo bindings in vkdt. the `layout(set = 0, binding = 2) uniform` will always be bound and contain the uint64_t addresses of the `ssbo` channel connectors in order of their connector indices (images/samplers will be skipped and not have a slot in this uniform buffer).

you seem to have run into such limitations with storage images though?

niklasiivari:
The issue i faced was when trying to allocate memory for the per frame ssbo in the filmsim rewrite. Vkdt wanted to allocate device local non-host-visible memory but a compatible type was not available on my hardware.

PR #289, opened 2026-07-28
filmsim: align with spektrafilm

niklasiivari:
Bit of a bigger change this time. This pr rewrites the filmsim module to align it with latest changes in the spektrafilm project. Things are more complicated than before, but tried to make the pipeline structure clearer by restructuring a lot of it. 

Performance takes a bit of a hit, but nothing too major considering we need to evaluate a lot more math than before. 

Sorry in advance, I know this will be a pain to review. 

Note: includes changes that break old cfgs.


hanatos:
oh wow! one hell of a patch indeed. will look into it, certainly excited!

some quick comments:
the CAT16 chromatic adaptation and the gamut compression is also from recent spektrafilm? makes me sad to see matrices in all the nice spectral surroundings ;) i'd rather include some spectral white balancing (divide out the scene illuminant). that would of course depend on some sanitised camera -> rec2020 input device transform.

niklasiivari:
Yeah, all those are directly from spektrafilm, but now that you mention it, would certainly make sense to try doing those in spectral too! I guess the reason for why it's done this way is the film's/paper's illuminant calibrations, but can't now figure out why doing the adaptation in spectral domain wouldn't work just as well.

In general there are now quite many stages that do a sort of "white balance" and would perhaps make sense to think about a way to make it make more sense. I also added one more illuminant slider besides scan and scene from yours, which I found useful for simulating for example exposing a tungsten-balanced stock in daylight (set film  to Kodak 200T and film illuminant to 5500 for example, and get a nice blue cast :)

hanatos:
really nice work here. i'd want to read the code some more to understand what's going on. from a few tests i have to say i like this a lot. many interesting new parameters. some comments:

- grain has global.hash in the seed so it still works nicely with animations, awesome!
- is couplers on and at 1.0 the default in spektrafilm too? seems a bit on the aggressive side. maybe 0.5 would lead to more balanced results?
- great to have all the illuminants to play with all the way at the bottom :)
- maybe the module would now profit from long names on the sliders, as did `OpenDRT`. there's the possibility to put a  [`params.en`](https://github.com/hanatos/vkdt/blob/master/src/pipe/modules/OpenDRT/params.en) file next to the `params.ui` file. it just holds pairs of `<token>:<long name>` in every line.
- does the halation `radius` do anything? can't spot the difference. maybe that affects a blur kernel which is implemented as multi-pass thing and then doesn't rebuild the graph? have to read the code.
- probably more to come :)

niklasiivari:
Some answers to the questions: 

- couplers 1.0 is the spektrafilm default too, but i agree it can be quite strong. 0.5 might make sense! Especially with positive stock 1.0 looks quite off. If we change the default should probably also regenerate wb tables as those are now also calibrated to couplers at 1. 
- yes, long names would surely help with ux :)
- just tested halation radius and seems to work for me, but the effect is quite subtle at these defaults anyways, if i set halation amount to say 100, there is quite a difference between for example radius 10 and 50. Yes its the blur kernel, but the rebuild should be handled ok!

niklasiivari:
Also a quick side note here not related to this pr, the double buffer changes seem to cause the gui to need to wait for the image computation every time, so slider changes + a slow graph make the ui stall once again

hanatos:
okay thanks for the comment. i'll revert yet again and figure this out with more time.

niklasiivari:
Just updating here that I actually started working on a fully spectral version! The idea I got is such that we have an alternative data format for representing the spectral data instead of rgb, produced by a module we can place at the place of the colour module. It evaluates a different-shaped lut with the idea of mapping camera-rgb > spectral instead of camera-rgb > rec2020. Then we can connect modules after this that work directly on spectral data. For example, we can directly simulate stuff like lenses, chromatic aberration (makes me think, could we figure out something for lens correction here with an inverse lens transformation!), or simply creative (or corrective) scene relighting.  

This works quite well with the graph system in vkdt because in theory we can then have both spectral and rgb sections in the graph with suitable transform modules in between. Will try to get something working by tomorrow for you to look at as well!

niklasiivari:
And here we go, thought not to open a PR yet due to how experimental this is at this point: https://github.com/niklasiivari/vkdt/tree/spectral

hanatos:
> Also a quick side note here not related to this pr, the double buffer changes seem to cause the gui to need to wait for the image computation every time, so slider changes + a slow graph make the ui stall once again

could you check a468d3a9 on your end if that's better?

niklasiivari:
> > Also a quick side note here not related to this pr, the double buffer changes seem to cause the gui to need to wait for the image computation every time, so slider changes + a slow graph make the ui stall once again
> 
> could you check [a468d3a](https://github.com/hanatos/vkdt/commit/a468d3a9e03e7ca2bff433f5863f685d835a502b) on your end if that's better?

Yes, works well now!

hanatos:
i think this is really outstanding work. really nice and clean, well structured and so much fun to play with :)

i put one minor change on top just because the way the grain reacted to changes of size threw me off the illusion of working with real film scans.

PR #290, opened 2026-07-31
filmsim: switch film illuminant adaptation from cat16 to spectral

niklasiivari:
So figured out this would be better done this way based on the earlier discussion, and seems to work just fine! Doing the "adaptation" in the spectral domain also removes the need for gamut compression, so a nice simplification. 

edit: here is a picture, less blue image is from this new version, kodak 500T with film illuminant set to 5500

image

Thought about scanner illuminant too, but it seems it kind of loses its meaning and is basically locked to D65 if we want to have a sensible white point on the rgb coming out of filmsim if we were to remove the CAT adaptation there.


hanatos:
```
 8 files changed, 60 insertions(+), 165 deletions(-)
```
love it.

Issue #294, opened 2026-08-04
from lighttable, opening a photo in darkroom requires 3 clicks


qosch:
I'm assuming this bug is specific to my (quite old) laptop, otherwise I'd expect someone else would have reported it already:

When I'm in the darkroom and the photo under the cursor is not selected:
- one click will select the photo (expected)
- double click will also just select the photo (would expect that it opens in darkroom)

When the photo under the cursor is already selected
- one click will do nothing (expected)
- a double click will open the photo in darkroom (expected)

Do you observe the same behaviour? If yes, is this intended?

I can try to fix it (if there is anything to fix), but I'd first need to know if this is specific to my machine and what is the expected behaviour. I just double checked: my "expected behaviour" remarks are how e.g. the KDE file manager Dolphin does it.

After starting vkdt, I did wait until all thumbnails were generated to be sure it's not my laptop responding sluggishly.

lu-k:
I have had some moments, where I double clicked on an image in Lighttable and nothing happened, but had blamed my pentablet. After seeing your report I went and checked specifically and can't fully confirm the behavior that you are describing. ~~ In my case it seems I only run into this issue once right after starting the app. ~~ 
Here is what I found:
- Pentablet - the issue sometimes occurs
- Mouse - never
- Touchpad double tap - always
- Touchpad double click (no buttons but there is a click point) - never

This is on a Laptop with Pop-OS 24.04 (Ubuntu), Wayland, running the AppImage with pentablet.

hanatos:
hm, my mouse is also the `never` category it seems. also i'd often use arrows + enter or gamepad to navigate.

your patch in #295 looks reasonable to me. maybe something about order of events or not receiving the first click in a separate render dispatch. clearly safer after the pr.

i've shortened the logic of the patch a bit, i think it's equivalent. let me know if that works on your setups :)

lu-k:
Seems fixed, couldn't reproduce with the merged PR

hanatos:
thanks for confirming. closing for now.

qosch:
>     * Pentablet - the issue sometimes occurs
> 
>     * Mouse - never
> 
>     * Touchpad double tap - always
> 
>     * Touchpad double click (no buttons but there is a click point) - never

Thanks for testing!

The fact that this seems to be input device dependend makes me think I haven't gotten to the bottom oft this yet and my PR only fixes the symptoms. Might habe to take another look.

qosch:
[...]

PR #297, opened 2026-08-07
dr: show thumbnail placeholder while pipeline processes

qosch:
During testing of #294 I noticed that it is hard to tell if a double click was registered, because there is no immediate feedback: while the pipeline runs, the lighttable continues to be displayed (frozen). This state lasts for about 3s on my laptop.

[Here](https://discuss.pixls.us/t/installed-at-last-but-unusable/59658/4) , I read 

> Unfortunately your gpu seems to only expose one vulkan queue so you do not benefit from running compute asynchronously from the gui

So this issue might only be present/pronounced on this class of old and slow GPUs.

To fix the "no immediate reaction to user action" issue, I figured it would be best to enter the darkroom immediately, and show the thumbnail (enlarged) while the full res pipeline is still running. I tried to keep complexity and amount of code as low as possible.

Let me know if you find this feature valuable and if it behaves well on your (faster) systems. The most important things for features that only slow systems benefit from, is, that they don't hurt on faster systems (I'd say).

### DIFF
--- a/src/gui/render_darkroom.c
+++ b/src/gui/render_darkroom.c
@@ -411,7 +411,6 @@ void render_darkroom_full(char *filter_name, char *filter_inst)
 
 void render_darkroom()
 {
-  // double clock_beg = dt_time();
 
   int win_x = vkdt.state.center_x,  win_y = vkdt.state.center_y;
   int win_w = vkdt.state.center_wd, win_h = vkdt.state.center_ht - vkdt.wstate.dopesheet_view;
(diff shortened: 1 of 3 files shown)


hanatos:
thanks, i'll need to look at that on an integrated gpu to appreciate what's happening :)

src/gui/render_darkroom.c
```diff
@@ -450,13 +449,24 @@ void render_darkroom()
   if(nk_begin(&vkdt.ctx, "darkroom center", bounds, NK_WINDOW_NO_SCROLLBAR | (disabled ? NK_WINDOW_NO_INPUT : 0)))
   { // draw center view image:
     dt_node_t *out_main = dt_graph_get_display(&vkdt.graph_dev, dt_token("main"));
-    if(out_main)
+    int rdy = out_main && (vkdt.graph_res[display_frame] == VK_SUCCESS);
```
hanatos, line 452: ah, no, `out_main` is different to ready `rdy`. if there is no output, we don't want to place a nuklear widget at all and that's fine. if there is an output, but it's not ready because the gpu is still busy, i want to place the widget but with a dummy/no image.

the context is that nuklear uses hashes to identify active widgets. in particular one situation is when you are dragging a slider it will grab the cursor as long as the widget is active. now if on the way the hashes change, the nuklear widgets get confused and your mouse ungrabs.

this is super annoying but luckily doesn't happen much. the situation where it happens is when a slider collapses to zero, like denoising strength. that will trigger a pipeline rebuild because a different set of kernels has to be dispatched. that means *both* output double buffers will be invalidated. these are owned by the graph, not the ui. that means now the ui doesn't have an image to display for a short time.. and that triggers the situation outlined above, hence the `rdy` dance.

hanatos:
thinking about this maybe it would be better to go the other way around and every once in a while run a quick dispatch that downsizes the output image to thumbnail size, transferring ownership to the thumbnail cache? for additional credits this would already have updated the thumbnail when going back to lt mode. such a thumbnail could be displayed in case `!rdy` because it's owned by the cache and hence not cleaned up on pipeline reconfiguration.

the assumption that thumbnail and full image are same aspect ratio is slightly risky, people could have cropped/rotated/done whatever. the thumbnail is now only recreated when leaving darkroom mode.

src/gui/widget_image.h
```diff
@@ -436,3 +436,31 @@ dt_image(
   // now the controls:
   if(events) dt_image_events(ctx, w, hover, main, disp);
 }
+
+// Draw a thumbnail as placeholder while the full-res pipeline is still processing.
+// Simple fit-to-view — no zoom/pan interaction. Both thumbnail and full-res
```
hanatos, line 441: argh ai-generated long hyphen makes me feel insecure about code quality surrounding this comment :D

hanatos:
..drafted something to illustrate what i mean. not sure it's pretty, doesn't belong to gui.h nor to graph.h, but kinda both: https://github.com/hanatos/vkdt/compare/master...dbuf#diff-1c9cadae23cc0903037091b6773aa9ea79338e1404a28983ea4f28b5f3493ca1

hanatos:
see #299. not sure this solves the issue addressed here, but i think it might (and anyways has potential to do the thumbnail rendering etc).

PR #298, opened 2026-08-09
Some cleanup commits for what I contributed earlier

niklasiivari:
Just some small cleanup commits fixing the following:

- colour: temp slider range went all the way to 0 and could trigger accidental autotemp
- filmsim lut: was packing unread information
- filmsim: added film base density control to match paper's similar parameter
- filmsim: scatter was defaulted to 1.0 which produces quite a soft result which is usually not ideal. Lowered to a more balanced 0.2
- filmsim: some logic that was truly per-frame computable was still done per-workgroup, now moved to setup shader. 
- upsample: did not need a rgb_to_xyz overload anymore since film illuminant was moved to spectral domain. Removed. 
- And some smaller clarity or duplication fixes.


hanatos:
looks good to me, thanks!

how is the cyan=-1 case supposed to be triggered? the ui element is hidden by default.. should that replace the `wb.h` preprocessing?

niklasiivari:
> looks good to me, thanks!
> 
> how is the cyan=-1 case supposed to be triggered? the ui element is hidden by default.. should that replace the `wb.h` preprocessing?

That's a remnant from a time when the sliders were still visible iirc, that is, it's been there already before I've touched this. Should probably clean that up though as it doesn't make much sense anymore.

hanatos:
ah okay then i think i remember. it's the default value, and it triggers loading from `wb.h` precomputed filter values.  so i suppose that's still needed then, unless you completely remove the precomputed wb.

niklasiivari:
> unless you completely remove the precomputed wb.

Nope! Everything is still using precomputed tables

PR #301, opened 2026-08-13
crop: improve auto crop

niklasiivari:
This pr reworks the crop module's auto crop feature to deterministically find the largest crop given the parameters.

There is some new behaviour:

- `-1` is a sentinel for "keep original aspect ratio".
- auto crop respects the aspect ratio setting	
- auto crop chooses the crop option that is closest to the current set crop when there are multiple options with equal area.


hanatos:
awesome. works as advertised and is super accurate. keeps exactly the promised 2px filter support margin. if i had to nitpick, for close to 45 degree rotations, it will keep the margin to top and left, but not diagonally (don't think this will ever result in any visible filtering issue, also not sure 45 degree rotation is a frequent use case).

great to see a clear improvement in one of the more wonky places of the code base :)

PR #306, opened 2026-09-12
JPEG-XL Export

WFinster:
Adds an `o-jxl` module. And makes a few extra changes to the macOS build. This is just an output module, with fairly limited options. Only 3-channel RGB, no animations, and some primaries and arbitrary ICC profiles aren’t implemented. Mostly done by following examples in other modules and with limited C knowledge, so hopefully haven’t done anything too silly :)


WFinster:
And I wasn’t quite thinking of the commit history as a public-facing record, so it’s a bit messy.

src/pipe/modules/o-jxl/params.ui
WFinster, line 1: The effort slider currently shows up like this:

Image

With no labels. And I wasn’t sure how to add them.

src/pipe/modules/o-jxl/connectors
WFinster, line 1: Maybe this should be f32? I initially thought that the node graph only used f16 precision but some modules seem to output f32. Lossy will happily ingest f32, but lossless files would get bigger.

.github/workflows/nightly.yml
```diff
@@ -344,39 +346,67 @@ jobs:
...
-                # lib=$(echo $lib | awk "/${BREW_PREFIX//\//\\/}/{ gsub(/[[:space:]].*/, \"\", \$1); print \$1 }")
-                name=$(basename "$lib")
-                [[ "$name" = "" ]] && continue
-                echo "lib dependency $lib"
-                echo "name $name"
-                install_name_tool -change $lib @rpath/$name $exe
-                if [ ! -f "${APPDIR}/Contents/Frameworks/${name}" ]; then
-                  cp -L "$lib" "${APPDIR}/Contents/Frameworks/${name}"
-                  found=1
+          function install_dependencies {
```
WFinster, line 355: This is from the darktable packaging scripts. The change was necessary as so far all homebrew dependencies were linked by their absolute path, which is how the previous CI found them. But some homebrew libraries, including some of the `libjxl` ones, link to relative dependencies… for some reason. This function handles that, and anyway seems a lot more elegant than the "keep copying dependencies until the number stops increasing" approach.

.github/workflows/nightly.yml
```diff
@@ -13,7 +13,7 @@ permissions:
 
 jobs:
   AppImage:
-    if: github.repository == 'hanatos/vkdt'
+    if: github.repository == 'WFinster/vkdt'
```
hanatos, line 16: i may want to revert this change :)

bin/config.mk.defaults
```diff
@@ -11,6 +11,11 @@ VKDT_JPEG_CFLAGS=$(eval VKDT_JPEG_CFLAGS := $$(shell pkg-config --cflags libjpeg
 VKDT_JPEG_LDFLAGS=$(eval VKDT_JPEG_LDFLAGS := $$(shell pkg-config --libs libjpeg))$(VKDT_JPEG_LDFLAGS)
 export VKDT_JPEG_CFLAGS VKDT_JPEG_LDFLAGS
 
+# Similar for finding libjxl
+VKDT_JXL_CFLAGS=$(eval VKDT_JXL_CFLAGS := $$(shell pkg-config --cflags libjxl libjxl_threads))$(VKDT_JXL_CFLAGS)
+VKDT_JXL_LDFLAGS=$(eval VKDT_JXL_LDFLAGS := $$(shell pkg-config --libs libjxl libjxl_threads))$(VKDT_JXL_LDFLAGS)
+export VKDT_JXL_CFLAGS VKDT_JXL_LDFLAGS
+
```
hanatos: no VKDT_USE_JXL flag to make it optional?

src/pipe/modules/o-jxl/connectors
hanatos: right. the pipeline would normally not provide f32 values. you can also make it '*' and then ask the connector for the format before interpreting the data pointer mapped from gpu memory.

src/pipe/modules/o-jxl/params.ui
```diff
@@ -0,0 +1,3 @@
+filename:filename
```
hanatos, line 1: why does github format this so broken? looks like the file has newline characters at the beginning of the line instead of at the end?

src/pipe/modules/o-jxl/params.ui
hanatos: ah that's just lazy programmer. didn't take much care to make this show up as in darkroom mode because it was unused so far. see 2b2ebce36.

src/pipe/modules/o-jxl/params.ui
WFinster: It shows up correctly in Firefox and Safari for me…
And the file itself seems correct locally (the newlines are LF, etc) —

Image

src/Makefile
```diff
@@ -17,6 +17,11 @@ ifeq ($(VKDT_USE_RAWINPUT), 0) # use no raw loading module
   MODULESXX:=$(filter-out i-raw,$(MODULESXX))
   MODULES:=$(filter-out i-raw,$(MODULES))
 endif
+ifneq ($(VKDT_USE_JXL), 1)
+  MODULES:=$(filter-out o-jxl,$(MODULES))
+  READMES:=$(filter-out o-jxl,$(READMES))
+  FLATMK:=$(filter-out o-jxl,$(FLATMK))
+endif
```
WFinster, line 24: The module still shows up as an export format in the GUI. But the same happens when disabling `o-vid` for example so I assume that’s something else.

bin/config.mk.defaults
WFinster: I think I just forgot, but also didn’t understand the Makefiles particularly well so was just following the jpg module’s example. I have given it an attempt now :)

src/Makefile
hanatos: ah good catch! the offender is here: https://github.com/hanatos/vkdt/blob/master/src/gui/widget_export.h#L119  might have to think about a good way to deal with this. maybe have to install the module directories one by one and omit the ones that aren't actually built (i.e. change the Makefile).

.github/workflows/nightly.yml
hanatos: i don't know nearly enough about macintonsh computers to judge this one :) i'll just go with your new version until someone can point out that it doesn't work :)

hanatos:
okay, thanks for the pull request and all the cleanups! the module itself may not be polished/full featured, but it does export jxl files and is wired correctly. i'd say this is ready to merge, we can cleanup/add features later as we go.

i added a few more commits about docs and made the jxl module optional with detection during build.

